PR #2, opened 2021-04-06
Strawman PR for Vaccine Spotter

rajbot:
Each runner should be supplied with an output-dir. The output-dir parameter can be a gs:// url.

The top level `runners/CA/vaccinespotter/run.py` script calls `fetch.py`, `parse.py`, then `normalize.py`. These can be called independently.

To scrape vaccinespotter for a different state, you would copy the four files in `runners/CA/vaccinespotter/` to a new directory, then change the `state = XX` line at the top of `run.py`. 

Scraping is started via `runners/XX/<source>/run.py`, and the actual ingestion code lives in `ingestors/<source>.py`.

Example output:
```bash
python3 runners/CA/vaccinespotter/run.py --output-dir out/CA/vaccinespotter

04/06/2021 23:22:40 INFO:run:Starting Vaccine Spotter CA crawler
04/06/2021 23:22:41 INFO:vaccinespotter:Fetching https://www.vaccinespotter.org/api/v0/states/CA.json
04/06/2021 23:22:41 INFO:vaccinespotter:Writing to out/CA/vaccinespotter/CA.geojson
04/06/2021 23:22:41 INFO:vaccinespotter:Parsing out/CA/vaccinespotter/CA.geojson
04/06/2021 23:22:41 INFO:vaccinespotter:Writing out/CA/vaccinespotter/CA.ndjson
04/06/2021 23:22:42 INFO:vaccinespotter:Normalizing out/CA/vaccinespotter/CA.ndjson to out/CA/vaccinespotter/CA.normalized.ndjson
```

First line of ndjson output (the full json schema isn't supported yet):
```
$ head -1  out/CA/vaccinespotter/CA.normalized.ndjson |jq
{
  "id": "vaccinespotter:7382066",
  "name": "Rite Aid",
  "address": {
    "street1": "1501 West Main Street",
    "city": "El Centro",
    "state": "CA",
    "zip": "92243"
  },
  "location": {
    "latitude": 32.7925,
    "longitude": -115.57
  },
  "booking": {
    "website": "https://www.riteaid.com/pharmacy/covid-qualifier"
  },
  "availability": {
    "appointments": false
  },
  "parent_organization": {
    "id": "rite_aid",
    "name": "Rite Aid"
  },
  "links": [
    {
      "authority": "rite_aid",
      "id": "5680"
    }
  ],
  "fetched_at": "2021-04-07T08:09:16.438145+00:00",
  "published_at": "2021-04-07T08:07:19.991+00:00",
  "sources": [
    {
      "source": "vaccinespotter",
      "id": "vaccinespotter:7382066",
      "fetched_from_uri": "https://www.vaccinespotter.org/api/v0/states/CA.json",
      "fetched_at": "2021-04-07T08:09:16.438145+00:00",
      "published_at": "2021-04-07T08:07:19.991+00:00",
      "data": {
        ...
      }
    }
  ]
}

```


jsonschema.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+
+from collections import namedtuple
+
+Location = namedtuple(
```
bryanculbertson, line 5: I think it would be useful to use the typed version of https://docs.python.org/3/library/typing.html#typing.NamedTuple for this.

Or, something with data marshaling built-in like:
- Pydantic: https://pydantic-docs.helpmanual.io/
- Attrs: https://www.attrs.org/en/stable/examples.html

bryanculbertson, line 5: I think we should name this "Clinic" instead of location, to separate it from the concept of the geo location

runners/CA/vaccinespotter.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
```
bryanculbertson, line 1: I would like the ability to run the 3 stages of fetch, parse, and normalize separately, even if they are run at once for a normal production run.

Perhaps something like:
1. `runners/CA/vaccinespotter/fetch.py` that conditionally fetchs the url and uploads to S3.
2. `runners/CA/vaccinespotter/parse.py` that fetchs the raw blob stored in S3, parses it into a raw source record in ndjson format, and uploads that to S3
3.  `runners/CA/vaccinespotter/normalize.py` that fetchs the ndjson stored in S3, normalizes it to match the schema I proposed, and uploads it to S3.
4. `runners/CA/vaccinespotter/run.sh` that runs each runner in succession.

obra:
@rajbot - The URL you've got in there may want to live in the booking: section of the json?

"booking: {
    phone: str as (###) ###-###,
    website: str,
    email: str,
    other: str,
  },
"

jsonschema.py
rajbot: We use Clinic for a specific kind of location type (vs Pharamacy, School, Nursing Home, etc)

bryanculbertson: Gotcha. What about `facility` then as the generic term? Or the even more generic `place`?

bryanculbertson: If these are called `Locations` elsewhere within VaccinateCA then I'll accept the ambiguity :)

rajbot: OK, I'll keep it Location, but am happy with either Facility or Place as well

runners/CA/vaccinespotter.py
rajbot: Refactored into four scripts as suggested

jsonschema.py
rajbot: Switched to typing.NamedTuple

ingestors/jsonschema.py
```diff
@@ -0,0 +1,115 @@
...
+            "latitude": loc.latitude,
+            "longitude": loc.longitude,
+        },
+        "contact": {
+            "phone": [loc.phone],
+            "website": [loc.website],
+            "email": [loc.email],
+            "other": [loc.contact_other],
+        },
+        "booking": {
```
obra, line 82: should the schema for contact and booking look identical? Seems like it may save headaches. don't feel strongly.

ingestors/jsonschema.py
```diff
@@ -0,0 +1,115 @@
...
+
+"""
+This module provides a flat Location structure so scraper authors do not need
+to understand the normalized ndjson spec:
+
+https://docs.google.com/document/d/1qxABDlRep76llrXhgwdwA6Ji3Jub-MftRvustk12XHc
+"""
+
+
+class Location(NamedTuple):
```
bryanculbertson, line 13: This looks like a schema defintion for a `VaccineSpotter` location, perhaps we should call it `VaccineSpotterLocation`?

ingestors/jsonschema.py
bryanculbertson: Yeah, this schema is awkward. I am trying to accomplish multiple things:
1. Provide a space for collecting multiple contact info if they exist.
2. Also provide a space for label something that we know is booking information as such.

Perhaps a better schema would be:
```
"contact": [{
  "contact_type": "booking",
  "phone": "(111) 222-3333",
  "website": "https://book.the.vaccine",
  "email": "vaccine@bookings.com",
  "other": "Book in person too",
}, ...]
```

ingestors/vaccinespotter.py
```diff
@@ -0,0 +1,89 @@
...
+
+import httpx
+from smart_open import open
+
+from . import header_dict, jsonschema
+
+logger = logging.getLogger("vaccinespotter")
+
+
+def fetch(state, geojson_file):
```
bryanculbertson, line 15: This is a fine first pass, but I expect we can make these generic, and then re-use them for different sources.

For example:
1. `fetchers/http.py`: With this method to call a url and dump the results to a file.
2. `parsers/geojson.py`: Converts a geojson file to an ndjson line per feature.
3. `normalizers/vaccinespotter.py`: Converts a vaccinespotter ndjson file into a VaccinateCA Locations schema ndjson file

runners/CA/vaccinespotter/fetch.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
```
bryanculbertson, line 1: Why a runner per-state, and instead of one runner for all states?

bryanculbertson (approved):
Approving this for merge, but I think there are still some high level concepts to work our afterwards

PR #7, opened 2021-04-10
Finish implementation of sf_gov example

rajbot:
Finish implementation of sf_gov example, as this is the simplest example we can provide.

The runner now takes an `--output-dir` parameter:
```
poetry run vaccine-feed-ingest/run.py all-stages --output-dir=<directory>
```

This directory should exist and be empty. The runner will create subdirectories for each site fetcher.

The cli parameters for fetch / parse / normalize have changed. The new parameters are:

```sh
fetch.sh <data_dir>
parse.sh <data_dir> <ndjson_path>
normalize.sh <ndjson_path> <normlized_ndjson_path>
```

Support parse.py and normalize.py, to keep crawler volunteers from having to jump through unnecessary hoops.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,14 @@ Pipeline for ingesting nationwide feed of vaccine facilities
 
 ## Usage
 
+### Quick Setup For MacOS Homebrew users
+
+```sh
+brew install python3
+brew install poetry
+poetry install
+```
+
 ### Setup Environment Once
 
 1. Install required system deps (Ubuntu/Debian):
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -4,6 +4,14 @@ Pipeline for ingesting nationwide feed of vaccine facilities
 
 ## Usage
 
+### Quick Setup For MacOS Homebrew users
```
bryanculbertson, line 7: I tried to make the steps environment agnostic and I test them on my mac and linux box. What issues did you run into when you tried them?

The steps you proposed aren't quite right because they could install the wrong version of python, and they suggest the user change their system python to run our project which many users will be loath to do if they do other python work.

bryanculbertson: If you are trying to avoid `pyenv` install then we could add that as an optional by suggested step?

PR #12, opened 2021-04-11
Fix vaccine_feed_ingest package name

bryanculbertson:
I forgot that packages can't have dashes in them. This was confusing linting tools.

I fixed by changing dashes to underscores.


bryanculbertson:
@rajbot I am merging this without review because it is such a big changes to change the name of a subfolder and I want to minimize merge conflicts for subsequent PRs. Feel free to review, or suggest changes to this though, and we can change them later.

PR #13, opened 2021-04-11
Fix available sites command

bryanculbertson:
Use find_executable to check if the stage exists

### DIFF
--- a/vaccine_feed_ingest/run.py
+++ b/vaccine_feed_ingest/run.py
@@ -72,7 +72,6 @@ def _find_executeable(site_dir: pathlib.Path, cmd_name: str) -> Optional[pathlib
     cmds = list(site_dir.glob(f"{cmd_name}.*"))
 
     if not cmds:
-        logger.info("No %s cmd in %s to run.", cmd_name, str(site_dir))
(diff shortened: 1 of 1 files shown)


vaccine_feed_ingest/run.py
```diff
@@ -107,6 +108,7 @@ def _run_fetch(site_dir: pathlib.Path) -> None:
 def _run_parse(site_dir: pathlib.Path) -> None:
     parse_path = _find_executeable(site_dir, PARSE_CMD)
     if not parse_path:
+        logger.info("No parse cmd in %s to run.", str(site_dir))
         return
 
     with tempfile.TemporaryDirectory(f"_parse_{site_dir.name}") as tmp_str:
```
rajbot, line 114: This will always pass a new empty directory to parse.sh. How will the parser get the data from fetch.sh?

bryanculbertson: I am fixing that in another PR. This PR is just to fix the available-sites command which I broke when I added support for fetch.sh and fetch.py

rajbot (approved):

PR #15, opened 2021-04-11
Support full pipeline run!

bryanculbertson:
`vaccine-feed-ingest all-stages` will run all stages for all sites and pass data around to each stage.

By default stores run data in local `out` directory, but allows specifying which directory to use with `--output-dir`

When a local directory is passes it used local file tools, when a gcs directrory is passed is uploads and downloads from gcs

### DIFF
--- a/vaccine_feed_ingest/run.py
+++ b/vaccine_feed_ingest/run.py
@@ -85,16 +85,16 @@ def _get_site_dirs(
     state: Optional[str], sites: Optional[Sequence[str]]
 ) -> Iterator[pathlib.Path]:
     """Return a site directory path, if it exists"""
-    if state is not None:
-        yield from _get_site_dirs_for_state(state)
-
-    elif sites is not None:
+    if sites:
         for site in sites:
             site_dir = _get_site_dir(site)
             if not site_dir:
                 continue
 
             yield site_dir
+
+    else:
+        yield from _get_site_dirs_for_state(state)
 
 
 def _find_executeable(
@@ -139,6 +139,9 @@ def _find_all_run_dirs(
     """Find latest stage output path"""
     stage_dir = base_output_dir / site / STAGE_OUTPUT_NAME[stage]
 
+    if not stage_dir.exists():
+        return
+
     for run_dir in sorted(stage_dir.iterdir(), reverse=True):
         if run_dir.name.startswith("_"):
             continue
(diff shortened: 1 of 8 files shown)


vaccine_feed_ingest/run.py
```diff
@@ -93,47 +127,163 @@ def _find_executeable(site_dir: pathlib.Path, cmd_name: str) -> Optional[pathlib
...
+
+        _copy_files(fetch_run_dir, parse_input_dir)
+
         subprocess.run(
-            [str(parse_path), str(tmp_dir / "output"), str(tmp_dir / "input")],
+            [str(parse_path), str(parse_output_dir), str(parse_input_dir)],
             check=True,
         )
 
+        _copy_files(parse_output_dir, parse_run_dir)
```
rajbot, line 245: Before this copy, _run_parse() should ensure at least one `*.parsed.ndjson` file was produced after `subprocess.run()`, and raise an Exception if one was not found.

vaccine_feed_ingest/run.py
```diff
@@ -93,47 +127,163 @@ def _find_executeable(site_dir: pathlib.Path, cmd_name: str) -> Optional[pathlib
...
+
+        _copy_files(parse_run_dir, normalize_input_dir)
+
         subprocess.run(
-            [str(normalize_path), str(tmp_dir / "output"), str(tmp_dir / "input")],
+            [str(normalize_path), normalize_output_dir, normalize_input_dir],
             check=True,
         )
 
+        _copy_files(normalize_output_dir, normalize_run_dir)
```
rajbot, line 285: Same comment as for `_run_parse()`. We should raise an exception if a *.normalized.ndjson file was not produced by subprocess.run().

vaccine_feed_ingest/run.py
```diff
@@ -93,47 +127,163 @@ def _find_executeable(site_dir: pathlib.Path, cmd_name: str) -> Optional[pathlib
...
+
     with tempfile.TemporaryDirectory(f"_fetch_{site_dir.name}") as tmp_str:
         tmp_dir = pathlib.Path(tmp_str)
+        fetch_output_dir = tmp_dir / "output"
+        fetch_output_dir.mkdir(parents=True, exist_ok=True)
+
+        subprocess.run([str(fetch_path), str(fetch_output_dir)], check=True)
 
-        subprocess.run([str(fetch_path), str(tmp_dir)], check=True)
+        _copy_files(fetch_output_dir, fetch_run_dir)
```
rajbot, line 207: When are the files saved from fetch.sh copied to GCS? Is it supposed to be done by _copy_files()?

vaccine_feed_ingest/run.py
```diff
@@ -93,47 +127,163 @@ def _find_executeable(site_dir: pathlib.Path, cmd_name: str) -> Optional[pathlib
...
+    fetch_run_dir = _generate_run_dir(
+        output_dir, site_dir.name, PipelineStage.FETCH, timestamp
+    )
+
     with tempfile.TemporaryDirectory(f"_fetch_{site_dir.name}") as tmp_str:
         tmp_dir = pathlib.Path(tmp_str)
+        fetch_output_dir = tmp_dir / "output"
+        fetch_output_dir.mkdir(parents=True, exist_ok=True)
+
+        subprocess.run([str(fetch_path), str(fetch_output_dir)], check=True)
```
rajbot, line 205: Occasionally we see scrapers fail to produce output due to upstream changes. We should check to make sure that fetch_output_dir is not empty, and raise an exception if it is

rajbot: Actually, raise an exception is the wrong thing, because we want the other scrapers to still run..

We should at least emit an logger.error and skip the parse phase if there is no output

vaccine_feed_ingest/run.py
bryanculbertson: Correct. Copy files will copy the files to the local `out` directory with running in development, but when running in production it will copy to GCS.

Depends on what the `--output-dir` is pointing to.

vaccine_feed_ingest/run.py
bryanculbertson: I will add code to check if the example didn't produce anything

rajbot (approved):

PR #18, opened 2021-04-14
Add Alabama fetcher

rajbot:
Example ArcGIS fetcher:

Run with:
```
mkdir out
poetry run vaccine-feed-ingest fetch --state=al --output-dir=out
```

ArcGIS json results will be copied to `out/`.

### DIFF
--- a/vaccine_feed_ingest/ingestors/arcgis.py
+++ b/vaccine_feed_ingest/ingestors/arcgis.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 
 import json
-import urllib3
 from os.path import join
+
+import urllib3
 
 http = urllib3.PoolManager()
 
--- a/vaccine_feed_ingest/ingestors/arcgis.py
+++ b/vaccine_feed_ingest/ingestors/arcgis.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+
+import json
+import urllib3
+from os.path import join
+
+http = urllib3.PoolManager()
+
+
+def get_count(query_url: str) -> int:
+    """
+    Get the total count of features in this ArcGIS feed.
+    This will be used for querying results in batches.
+    """
+
+    r = http.request(
+        "GET",
+        query_url,
+        fields={"where": "1=1", "returnCountOnly": "true", "f": "json"},
+    )
+    obj = json.loads(r.data.decode("utf-8"))
+    return obj["count"]
+
+
+def get_results(query_url: str, offset: int, batch_size: int, output_dir: str):
+    """ Fetch one batch of ArcGIS features from the query_url """
+
+    # Set Output Spatial reference to EPSG 4326 GPS coords
+    out_sr = "4326"
+
+    r = http.request(
+        "GET",
+        query_url,
+        fields={
+            "where": "1=1",
+            "outSR": out_sr,
+            "f": "json",
+            "outFields": "*",
+            "returnGeometry": "true",
+            "orderByFields": "objectId ASC",
+            "resultOffset": offset,
+            "resultRecordCount": batch_size,
+        },
+    )
+
+    # obj = json.loads(r.data.decode('utf-8'))
+
+    output_file = join(output_dir, f"{offset}.json")
+    with open(output_file, "wb") as fh:
+        print(f"Writing {output_file}")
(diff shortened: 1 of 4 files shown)


obra:
we can't get the native geojson feed out of arcgis for this one?

vaccine_feed_ingest/runners/al/arcgis/fetch.py
```diff
@@ -0,0 +1,20 @@
...
+runner_dir = state_dir.parent
+root_dir = runner_dir.parent
+sys.path.append(str(root_dir))
+from ingestors import arcgis  # noqa: E402
+
+output_dir = sys.argv[1]
+print(f"Starting AL scraper with {output_dir=}")
+
+
+query_url = """https://services7.arcgis.com/4RQmZZ0yaZkGR1zy/arcgis/rest/services/DriveThruWalkIn2_Public/FeatureServer/0/query"""  # noqa: E501
```
obra, line 18: Since they have multiple URLs we'll want to scrape, would we want to set a pattern of either having an array of URLs or of naming more descriptively?

vaccine_feed_ingest/ingestors/arcgis.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
obra, line 1: Since there are multiple arcgis endpoints, I might lean toward calling this arcgis_featureserver or something, but do not feel strongly.

vaccine_feed_ingest/runners/al/arcgis/fetch.py
rajbot: No longer using urls, since we switch to the arcgis package

vaccine_feed_ingest/ingestors/arcgis.py
```diff
@@ -0,0 +1,82 @@
...
+from arcgis import GIS
+
+http = urllib3.PoolManager()
+
+
+def fetch_geojson(
+    service_item_id: str,
+    output_dir: str,
+    selected_layers: Optional[Sequence[str]] = None,
+):
```
bryanculbertson, line 17: missing return type

```suggestion
) -> None:
```

vaccine_feed_ingest/ingestors/arcgis.py
```diff
@@ -0,0 +1,82 @@
...
+    r = http.request(
+        "GET",
+        query_url,
+        fields={"where": "1=1", "returnCountOnly": "true", "f": "json"},
+    )
+    obj = json.loads(r.data.decode("utf-8"))
+    return obj["count"]
+
+
+def get_results(query_url: str, offset: int, batch_size: int, output_dir: str):
```
bryanculbertson, line 48: missing return type

```suggestion
def get_results(query_url: str, offset: int, batch_size: int, output_dir: str) -> None:
```

vaccine_feed_ingest/ingestors/arcgis.py
```diff
@@ -0,0 +1,82 @@
...
+        },
+    )
+
+    output_file = join(output_dir, f"{offset}.json")
+    with open(output_file, "wb") as fh:
+        print(f"Writing {output_file}")
+        fh.write(r.data)
+
+
+def fetch(query_url: str, output_dir: str, batch_size=50):
```
bryanculbertson, line 75: Missing parameter type and return type

```suggestion
def fetch(query_url: str, output_dir: str, batch_size: int = 50) -> None:
```

vaccine_feed_ingest/runners/al/arcgis/fetch.py
```diff
@@ -0,0 +1,24 @@
...
+
+import pathlib
+import sys
+
+# import arcgis ingestor
+site_dir = pathlib.Path(__file__).parent
+state_dir = site_dir.parent
+runner_dir = state_dir.parent
+root_dir = runner_dir.parent
+sys.path.append(str(root_dir))
```
bryanculbertson, line 11: I'll see if I can so this in a better way in a later PR. I think if I pass `PYTHONPATH=vaccine-feed-ingest` when I call the script then the script authors won't need to setup the import path manually.

bryanculbertson (approved):

PR #24, opened 2021-04-15
chore: fetch MD ArcGIS

eliblock:
Fetch Maryland's information using the [ArcGIS runner](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/arcgis-runner).

### DIFF
--- a/vaccine_feed_ingest/runners/md/arcgis/fetch.py
+++ b/vaccine_feed_ingest/runners/md/arcgis/fetch.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+
+import pathlib
+import sys
+
+# import arcgis ingestor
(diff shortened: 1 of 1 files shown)


obra:
minor style request quest for these PRs: mentioned how many results we get and show an example of one record's worth of json￼.

obra:
given the shape of these, I wonder if the arcgis fetches could just be a yaml file listing the service ids and tab names, so there's zero custom code or boilerplate ￼

eliblock:
@obra - will do for future PRs, and if the pattern holds for a couple more states, I agree we should consider removing the common code (we might need to adjust our runner to pick up the state's ArcGIS from the YAML in addition to the per-state scripts).

PR #26, opened 2021-04-16
chore: shared fetch.py with yml config for arcgis

eliblock:
ArcGIS feeds have a repeatable `fetch` pattern, and @obra observed that
we could make it easier to add new feeds by re-using a single runner.

Augment `vaccine-feed-ingest` to accept `.yml` configuration files if an
executable is not provided (for the `fetch` stage), and use that config
to run a shared ArcGIS scraper. This setup is designed so that the
`vaccine-feed-ingest` cli commands continue to operate as expected even
when a config file is found rather than an executable runner. Additionally,
while ArcGIS is the only shared scraper at the moment, this setup should
be able to support more shared scrapers with minimal refactoring.

Then, convert existing ArcGIS feeds to use the new implementation.

A sample `fetch.yml` is as follows. This file would be placed at
`vaccine_feed_ingest/runners/md/arcgis/fetch.yml` (but this example
is not MD's actual config).
```yml
---
state: md
arcgis:
- id: d677f142234648a1a41b84d94df8e134
  layer_names:
  - Mass Vaccination Sites
  - Pharmacy Vaccination Sites
- id: 9f23e1c3b5c54198ab61d2f729cb787d
  layer_names:
  - FederalPartners
```


rajbot (approved):

bryanculbertson (approved):
I like this idea!

I think it has further potential to abstract in a way to that supports other types of shared runners. That can happen in a future PR when it comes up.

PR #29, opened 2021-04-17
chore: return all records in ArcGIS fetch

eliblock:
Per [ArcGis's python documentation](https://developers.arcgis.com/python/api-reference/arcgis.features.toc.html#arcgis.features.FeatureLayer.query), setting `return_all_records = True` on a `query` causes:

> When True, the query operation will call the service until all records that satisfy the where_clause are returned.

It seems desirable to fetch all data - not just the first page - so add this parameter.


obra (approved):
I love this change

PR #32, opened 2021-04-17
feat!: namespace output and temp dirs

eliblock:
Currently runners are namespaced: `<state_or_us>/<site>/<stage>`, but our output directories are simply `<site>/<stage>` and our temporary working directories are simply `<stage>_<site>`. This means that site names must be globally unique to ensure that files end up in isolated directories.

As a result of this, non-unique site names (like `arcgis`) were being repeatedly parsed and normalized because the temp directories were shared.

Update `run.py` to:
* name temporary working directories `<stage>_<site_or_us>_<site>`
* name output directories `<state_or_us>/<site>/<stage>`


rajbot (approved):
LGTM

PR #34, opened 2021-04-18
chore: add normalized schema

eliblock:
- chore: add normalized schema data types

Codify discussed target schema into classes with type hints, using
pydantic for runtime validation and serialization.

Future work can improve these types (e.g., enforcing ISO format for the
date and time fields rather than accepting arbitrary strings), but this
should be a sufficient starting point.

- chore: normalize ak/arcgis using data types

A sample normalization stage using the above types.

### DIFF
--- a/vaccine_feed_ingest/schema/schema.py
+++ b/vaccine_feed_ingest/schema/schema.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python3
+
+from typing import List, Optional
+
+from pydantic import BaseModel
+
+
+class Address(BaseModel):
+    """
+    {
+        "street1": str,
+        "street2": str,
+        "city": str,
+        "state": str as state initial e.g. CA,
+        "zip": str,
+    },
+    """
+
+    street1: str
+    street2: Optional[str]
+    city: str
+    state: str
+    zip: str
+
+
+class LatLng(BaseModel):
+    """
+    {
+        "latitude": float,
+        "longitude": float,
+    },
+    """
+
+    latitude: float
+    longitude: float
+
+
+class Contact(BaseModel):
+    """
+    {
+        "contact_type": str as contact type enum e.g. booking,
+        "phone": str as (###) ###-###,
+        "website": str,
+        "email": str,
+        "other": str,
+    }
+    """
+
+    contact_type: Optional[str]
+    phone: Optional[str]
+    website: Optional[str]
+    email: Optional[str]
+    other: Optional[str]
+
+
+class OpenDate(BaseModel):
+    """
+    {
+        "opens": str as iso8601 date,
+        "closes": str as iso8601 date,
+    }
+    """
+
+    opens: Optional[str]
+    closes: Optional[str]
+
+
+class OpenHour(BaseModel):
+    """
+    {
+        "day": str as day of week enum e.g. monday,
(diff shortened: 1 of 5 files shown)


vaccine_feed_ingest/schema/schema.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+
+    opens: Optional[str]
+    closes: Optional[str]
+
+
+class OpenHour(BaseModel):
+    """
+    {
+        "day": str as day of week enum e.g. monday,
```
rajbot, line 71: Can this be a day range, like M-F?
Can this be optional, to indicate same hours every day?

vaccine_feed_ingest/schema/schema.py
```diff
@@ -0,0 +1,182 @@
...
+        "opens": str as hh:mm:ss,
+        "closes": str as hh:mm:ss,
```
rajbot, line 73: Are seconds required here for open/close times?

vaccine_feed_ingest/schema/schema.py
nschiefer: To the extent possible, I would _really_ like to have business hours in a structured form (so that we can direct phone bankers/call centres to open locations, especially given time zones). I am okay with deciding that the normalization happens somewhere else, but it does need to live somewhere.

nschiefer:
This is not blocking at all, but I would be interested in pulling these data objects out into some kind of library so that Vial can use them as well. I am currently working on a source location -> canonical location converter that might want to live in this library, too.

vaccine_feed_ingest/schema/schema.py
obra: No

vaccine_feed_ingest/schema/schema.py
obra: At least at the vial level, we need to be able to ask the question "is this open now"? 

We won't have the data everywhere, but we should be representing it as spans of time that the location is open. The spec @bryanculbertson put together had this normalized to an array of "day of week, opens at, closes at" tuples. This neatly solves the problem of a pharmacy being open from 9-noon and then 1-6 on sundays.  

At the same time, we get a lot of really bad opening hours data and it may make sense to have an escape hatch to include an unstructured "opening hours in prose" field of some kind, so we don't have to implement all of RFC2445's recurrence calculation engine?

I'd much rather we be able to get data into the database and then iterate on parsing out opening hours later.

vaccine_feed_ingest/schema/schema.py
bryanculbertson: We can remove seconds

eliblock: Will open a follow-up to remove seconds from this comment (since it is a shift from the original spec).

vaccine_feed_ingest/schema/schema.py
eliblock: Per discussion on Discord:
* Current schema is based on https://schema.org/OpeningHoursSpecification.
* May also want to track "hard-to-normalize free text hours," but this will be added later if desired.

Leaving as-is.

rajbot (approved):

PR #37, opened 2021-04-19
Add `load-to-vial` command

bryanculbertson:
Add a command to load normalized locations to VIAL

Right now this loads all source locations as new. Matching logic will be added in a followup PR.

```sh
poetry run vaccine-feed-ingest load-to-vial ca/sf_gov
```


rajbot (approved):
LGTM! Left a note about isort..

PR #74, opened 2021-04-23
A first pass at a setup.sh script for macOS

obra:


setup.sh
```diff
@@ -0,0 +1,82 @@
...
+    fi
+    echo "Installing all of our python dependencies using Poetry."
+
+    poetry install --extras lint
+
+    echo "Install done."
+    echo ""
+    echo "Try this command next:"
+    echo ""
+    echo "poetry run vaccine-feed-ingest --help"
```
eliblock, line 38: Can/should we simply run this as the last command?

bryanculbertson (approved):

setup.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+
+maybe_install() {
+
+    exists=$(which "$1")
+    if [ "xx$exists" == "xx" ]; then
```
eliblock, line 6: I've previously used code like the following (I didn't actually run this so it might not work) to do this sort of thing.

```sh
if [ -x "$(which $1)" ]; then
    ... # installed case
else
    ... # not installed case
end
```

That said, I love how readable it is the way this is written!

eliblock (approved):
🚀 faster onboarding!

setup.sh
```diff
@@ -0,0 +1,82 @@
...
+        echo "$2 doesn't seem to be installed locally, but I can do it for you."
+        echo ""
+        echo "Hit Control-C if you don't want me to install $2 for you"
```
eliblock, line 9: Should this say "or hit enter to continue," or something similar?

setup.sh
obra: Yeah. This is "old habits die hard" combined with "shellcheck made demands"

setup.sh
obra: it probably should, yes.

PR #84, opened 2021-04-23
Add Fetch + Parse steps for Mass.

### LINKED ISSUE #69: fetch vaxfinder.mass.gov from Massachusetts
[learn our pipeline: fetch](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/Runner-pipeline-stages#fetch)

Fetch data from: https://vaxfinder.mass.gov (note multiple pages of results)

Put your script in a file named: `ma/vaxfinder_gov/fetch.sh` (`.py`, etc. are also okay)

Store it (without processing) in a new file created in the directory passed as the first argument (`sys.argv[1]`).

Check the wiki to learn more about the purpose of the fetch stage and how to get set up for development!

### Tips

1. While working on your code, run it at any point:
    ```sh
    poetry run vaccine-feed-ingest fetch ma/vaxfinder_gov
    ```

Deon:
Resolves #69 and #70.

This PR handles the fetch + parse steps for MA. 

Since we're not using the availability data at the moment, I decided to use their xlsx source as it's easier to get started with - ie. no pagination required, dates are better formatted. 

See slack thread: https://vaccine-feeds.slack.com/archives/G01PX9KH34H/p1619204038086500?thread_ts=1619203431.085900&cid=G01PX9KH34H

Sample data:
```
{"name": "Haverhill: Walgreens (River St.)", "address": "800 River St., Haverhill, MA 01832"}
{"name": "Haverhill: Walgreens (South Main St.)", "address": "68 South Main St, Haverhill, MA 01835"}
{"name": "Hingham: CVS (Lincoln St.)", "address": "405 Lincoln St, Hingham, MA 02043"}
{"name": "Hingham: CVS (South Street)", "address": "86 South Street, Hingham, MA 02043"}
```

### DIFF
--- a/vaccine_feed_ingest/runners/ma/parse.py
+++ b/vaccine_feed_ingest/runners/ma/parse.py
@@ -9,18 +9,17 @@ import pandas
 output_dir = pathlib.Path(sys.argv[1])
 input_dir = pathlib.Path(sys.argv[2])
 
-csv_filepaths = input_dir.glob("*.xlsx")
+xlsx_filepath = input_dir / "ma.xlsx"
 
-for in_filepath in csv_filepaths:
-    locations = []
-    data = pandas.read_excel(in_filepath, engine="openpyxl")
+locations = []
+data = pandas.read_excel(xlsx_filepath, engine="openpyxl")
 
-    for row in data.itertuples():
-        locations.append(dict(name=row[1], address=row[3]))
+for row in data.itertuples():
+    locations.append(dict(name=row[1], address=row[3]))
 
-    out_filepath = output_dir + "/data.parsed.ndjson"
+out_filepath = output_dir / "data.parsed.ndjson"
 
-    with out_filepath.open("w") as fout:
-        for location in locations:
-            json.dump(location, fout)
-            fout.write("\n")
+with out_filepath.open("w") as fout:
+    for location in locations:
+        json.dump(location, fout)
+        fout.write("\n")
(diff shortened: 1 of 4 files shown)


vaccine_feed_ingest/runners/ma/parse.py
```diff
@@ -0,0 +1,27 @@
...
+
+for in_filepath in csv_filepaths:
+    locations = []
+    data = pandas.read_excel(in_filepath, engine="openpyxl")
+
+    for row in data.itertuples():
+        locations.append(dict(name=row[1], address=row[3]))
+
+    filename = in_filepath.name.split(".", maxsplit=1)[0]
+    out_filepath = output_dir / f"{filename}.parsed.ndjson"
```
rajbot, line 22: Could you make this always be `data.parsed.ndjson` for consistency with our other parsers?

vaccine_feed_ingest/runners/ma/parse.py
```diff
@@ -0,0 +1,26 @@
...
+for in_filepath in csv_filepaths:
+    locations = []
+    data = pandas.read_excel(in_filepath, engine="openpyxl")
+
+    for row in data.itertuples():
+        locations.append(dict(name=row[1], address=row[3]))
+
+    out_filepath = output_dir + "/data.parsed.ndjson"
+
+    with out_filepath.open("w") as fout:
```
rajbot, line 23: Instead of writing an output file for each file (which might overwrite here), we would records from all csv files combined into a single data.parsed.ndjson file. Could you move the fout open to the outer loop, and write each record into the same file?

Deon: Ah. We only have the one file right now, so I think it'd be cleaner just to read the one in this case. I'll make the change and that would mean we don't need the outer loop.

rajbot (approved):

PR #86, opened 2021-04-23
Add us/vaccinespotter_org parser

juleea:


vaccine_feed_ingest/runners/us/vaccinespotter_org/parse.py
```diff
@@ -0,0 +1,22 @@
...
+input_dir = pathlib.Path(sys.argv[2])
+
+json_filepaths = input_dir.glob("*.geojson")
+
+for in_filepath in json_filepaths:
+    with in_filepath.open() as fin:
+        raw_json = json.load(fin)
+
+    filename = in_filepath.name.split(".", maxsplit=1)[0]
+    out_filepath = output_dir / f"{filename}.parsed.ndjson"
```
rajbot, line 17: We would like the output file to be a single `data.parsed.ndjson` file that contains every site. Could you move the fout open to the outer loop, and safe all records into a single file?

rajbot (approved):

eliblock:
Fixes #47

PR #119, opened 2021-04-24
add fetch.yml for GISCorps national locations for vaccination sites

MoralCode:
GIScorps is a group of volunteers familiar with GIS systems that have been compiling a national dataset of testing (and later vaccination) locations for a few months already. Might be worth reaching out to them. Let me know and I can do an introduction with the person I've been in email contact with.

This data is likely to overlap quite a lot with existing data you may have and I'm sure they would also like to incorporate any new or modified data back into their own dataset for the various organizations that use that data. There are some details and links to their bulk-import spreadsheet and schemas at https://github.com/VacFind/GeoCompare/issues/7 and also in https://github.com/VacFind/GeoCompare/wiki/GISCorps-Output-Format

### DIFF
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/fetch.yml
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/fetch.yml
@@ -1,0 +1,7 @@
+---
+state: us
+arcgis:
+# https://covid-19-giscorps.hub.arcgis.com/app/3b4bd11928ec488281b0280d4d45533a
+- id: c50a1a352e944a66aed98e61952051ef
+  layer_names:
+  - Testing Locations # this seems strange that the layer is called testing locations but the data seems accurate


eliblock (approved):
Thank you! I opened #120 to parse this data (you can use the shared parser!) and #121 to normalize it.

The match-and-load stage of our pipeline is designed to handle de-duping datasets, it's perfectly okay to get overlapping data from multiple sources!

MoralCode:
@eliblock would it also be possible to (eventually) add some kind of mechanism after dedup that can allow data modified from/not present in a given dataset (like this giscorps one) to be formatted in a way that can be contributed back to that dataset too? this way data can be shared across other projects both to benefit the community and to effectively act as a redundant backup that can be re-pulled in using this script should one database get lost

This is probably more of a long term thing but just wanted to mention it

eliblock:
@MoralCode great question! I think about this in two ways:

* _In general_, I want to point towards [Vaccinate The State's about-us page](http://vaccinatethestates.com/about-us), which talks about the org's goals to collect data and then share the aggregated form back out to the public (via the site) as well as other publishers and public health initiatives.

* As far as the specific details of the question, I know some members of Vaccinate The State's team have already chatted with the folks at GISCorps. My hope is that those teams should be able to work out data sharing as needed!

obra:
With this dataset, I want to make sure we’re not also getting test-only locations.

Issue #85, opened 2021-04-23
Build regex to produce links from CVS store IDs


nschiefer:
In `utils/normalize.py`, we have a function `def provider_id_from_name(name: str) -> Optional[Tuple[str, str]]:` that tries to pull out retail pharmacy store IDs from names. This is a very reliable way to match.

We don't currently have one for CVS, and it would be great to have.

cdjj:
Pull request coming in a moment.

cdjj:
We can close this now that the PR's in.

PR #135, opened 2021-04-24
Add CVS regex.

cdjj:
Solves #85.

### DIFF
--- a/vaccine_feed_ingest/utils/normalize.py
+++ b/vaccine_feed_ingest/utils/normalize.py
@@ -33,5 +33,8 @@ def provider_id_from_name(name: str) -> Optional[Tuple[str, str]]:
     m = re.search(r"WALMART PHARMACY 10-(\d+)", name, re.I)
     if m:
         return "walmart", m.group(1)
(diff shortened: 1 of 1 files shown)


cdjj:
Incidentally, I've got a script that scrapes CVS for which cities have available appointments; would that be useful?

cdjj:
Sure thing --- I've extended it to match variants on CVS Pharmacy.

Re scraper, is there anywhere you'd like me to put it in the code base?  It's not currently a great fit for the fetchers, since it's just listing which cities have availability, not which stores.

cdjj:
Good catch (pun intended), fixed.

cdjj:
FWIW, I just spent a couple minutes poking at CVS's web flow.  It looks like it should be pretty straightforward to build a real fetcher for it (i.e. give it a ZIP code and get locations with availability), I'll try doing that later today.

obra:
we get cvs info nationally from vaccinespotter, so this may not be worth a ton of your time.

cdjj:
ACK.

cdjj:
Done.

cdjj:
Closes #85 .

Issue #114, opened 2021-04-24
fetch arcgis from tn


eliblock:
[![learn our pipeline: fetch](https://img.shields.io/static/v1?label=learn%20our%20pipeline&message=fetch&style=social)](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/Runner-pipeline-stages#fetch)

Fetch data from: https://experience.arcgis.com/experience/885e479b688b4750837ba1d291b85aed

Put your script in a file named: `tn/arcgis/fetch.yml` (see how to use this file [on the wiki](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/ArcGIS-FeatureServer))

Store it (without processing) in a new file created in the directory passed as the first argument (`sys.argv[1]`).

Check the wiki to learn more about the purpose of the fetch stage and how to get set up for development!

### Tips

1. While working on your code, run it at any point:
    ```sh
    poetry run vaccine-feed-ingest fetch tn/arcgis
    ```

SansSkill:
I can take this one

PR #136, opened 2021-04-24
Add TN Fetcher

SansSkill:
Solves #114


eliblock (approved):
Thank you @SansSkill! Let us know if you have any feedback on the shared-fetch format that you used here.

PR #137, opened 2021-04-24
Add Dockerfile

mbacchi:
Add Dockerfile for Docker usage to avoid installing tools locally.


mbacchi:
I tried creating docs in the Wiki for this, but cannot push a branch because I don't have permission. The changes I had staged would be added to the `Manual-Development-Environment-Setup.md` file:

```
diff --git a/Manual-Development-Environment-Setup.md b/Manual-Development-Environment-Setup.md
index a40e0cd..86ffb6d 100644
--- a/Manual-Development-Environment-Setup.md
+++ b/Manual-Development-Environment-Setup.md
@@ -142,4 +142,29 @@ Once you have set up your development environment, see [running `vaccine-feed-in
     ```sh
     curl https://sdk.cloud.google.com | bash
     gcloud init
-    ```
\ No newline at end of file
+    ```
+
+## Setup for Running in Docker
+
+**NOTE:** This does not setup a Docker volume mount to access the resulting data
+files generated from the `vaccine-feed-ingest` commands outside the container.
+If you need to do this you will need to run the container with the `-v` volume
+mount option.
+
+1. Build the Docker image
+
+    ```sh
+    docker build -t vaccine-feed-ingest .
+    ```
+
+2. Start the docker container and get a shell prompt
+
+   ```sh
+   docker run -it vaccine-feed-ingest
+   ```
+
+3. Run vaccine-feed-ingest commands
+
+    ```sh
+    poetry run vaccine-feed-ingest fetch ca/sf_gov
+    ```
```

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.9.2-buster
+LABEL name=vaccine-feed-ingest
+
+RUN useradd -m vaccine && mkdir vaccine-feed-ingest && chown vaccine:vaccine vaccine-feed-ingest
+USER vaccine
+
+RUN git clone https://github.com/CAVaccineInventory/vaccine-feed-ingest.git
```
rajbot, line 7: Thank you for contributing a Dockerfile!

To allow this Dockerfile to work in forked repos, can you use `COPY` to add the project to the container, instead of cloning it from upstream?

mbacchi: Good point. Thanks.

obra: A pattern I find works better is to bind mount the source dir from the checkout. The user will have already cloned to get the docker file. letting them live edit code in their native editor provides much more pleasant dev experience.

mbacchi:
> One problem when testing is that if someone has already run `poetry install` on the host system, and then attempts to build the container, they might hit this error:
> 
> ```
> #10 2.705 The virtual environment found in /vaccine-feed-ingest/.venv seems to be broken.
> #10 2.706 Recreating virtualenv vaccine-feed-ingest in /vaccine-feed-ingest/.venv
> #10 2.792 
> #10 2.792   PermissionError
> #10 2.792 
> #10 2.792   [Errno 13] Permission denied: 'nbclassic.json'
> ```
> 
> The easiest fix I think is to delete the `.venv` directory after the `COPY ./` command., or chown everything to the `vaccine` user.
> 
> Also, if the `out` directory is copied into the container, running `poetry run vaccine-feed-ingest fetch ca/sf_gov` might result in this error:
> 
> ```
> PermissionError: [Errno 13] Permission denied: 'out/ca'
> ```
> 
> Making sure there is a clean `out` directory owned by the `vaccine` user would fix that permission error as well.
> 
> Other than that, this is great! Thank you!

I've encountered a similar `.venv` error when working on the COPY step, because I ran initial attempts with a `-v` docker volume mount. I removed it manually, but could certainly do it in the build. Regarding the `out` directory I tried mounting that too just now and get permission errors. Let me see what else I can try to work around that.

mbacchi:
> > I've encountered a similar `.venv` error when working on the COPY step, because I ran initial attempts with a `-v` docker volume mount. I removed it manually, but could certainly do it in the build. Regarding the `out` directory I tried mounting that too just now and get permission errors. Let me see what else I can try to work around that.
> 
> If you add `.venv` and `out` to a `.dockerignore` file, then create a fresh out directory in your Dockerfile, it should fix it.

I will do that now. I assumed you meant you wanted the host `out` directory to be mounted and used even if it had previous data in it. This method is much more straightforward.

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN rm -rf /vaccine-feed-ingest/.venv/ && mkdir /vaccine-feed-ingest/out && chown vaccine:vaccine /vaccine-feed-ingest/out
+
+USER vaccine
+
+WORKDIR /vaccine-feed-ingest
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
+ENV PATH "/home/vaccine/.poetry/bin:$PATH"
+RUN /home/vaccine/.poetry/bin/poetry install --extras lint
```
bryanculbertson, line 16: Since you are installing in a container, you probably don't also need a venv. You can turn off venv with:
```poetry config virtualenvs.create false```

```suggestion
RUN /home/vaccine/.poetry/bin/poetry config virtualenvs.create false && /home/vaccine/.poetry/bin/poetry install --extras lint --no-interaction --no-ansi
```

mbacchi: Newest commit removes the .venv

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM python:3.9.2-buster
+LABEL name=vaccine-feed-ingest
+
+RUN useradd -m vaccine && mkdir vaccine-feed-ingest && chown vaccine:vaccine vaccine-feed-ingest
+
+COPY ./ /vaccine-feed-ingest/
+RUN rm -rf /vaccine-feed-ingest/.venv/ && mkdir /vaccine-feed-ingest/out && \
```
bryanculbertson, line 7: Do you still need this line with the .dockerignore ignore these files?

bryanculbertson (approved):
Merging your Dockerfile and we can iterate on it from there if we need to

PR #144, opened 2021-04-24
Add normalizer for GISCorps national vaccine providers

### LINKED ISSUE #121: normalize GisCorps national data
[learn our pipeline: normalize](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/Runner-pipeline-stages#normalize)

Normalize existing [`.ndjson`](http://ndjson.org/) records into [the Vaccinate The States data schema](https://github.com/CAVaccineInventory/vaccine-feed-ingest/blob/main/vaccine_feed_ingest/schema/schema.py).

Read all files in the directory passed as the second argument (`sys.argv[2]`), transform each record into the normalized schema (filling in as much detail as possible), then output them to new files in the directory passed as the first argument (`sys.argv[1]`).

Check the wiki to learn more about the purpose of the normalize stage and how to get set up for development!

### Tips

1. Fetch and parse data for this site before you start developing:
    ```sh
    poetry run vaccine-feed-ingest fetch us/giscorps_vaccine_providers
    poetry run vaccine-feed-ingest parse us/giscorps_vaccine_providers
    ```

1. While working on your code, run it at any point:
    ```sh
    poetry run vaccine-feed-ingest normalize us/giscorps_vaccine_providers
    ```

### Example
[Parsed files for `ak/arcgis`](https://github.com/CAVaccineInventory/vaccine-feed-ingest-results/tree/main/ak/arcgis/parsed) are converted to [normalized files](https://github.com/CAVaccineInventory/vaccine-feed-ingest-results/tree/main/ak/arcgis/normalized)

MoralCode:
Work in progress. Current progress documented in the linked issue

fixes #121

### DIFF
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -166,6 +166,20 @@ def _get_published_at(site: dict) -> Optional[str]:
     return None
 
 
+
+def try_get_list(lis, index, default=None):
+    if lis is None:
+        return default
+
+    try:
+        value = lis[index]
+        if value == "none":
+            logger.warn("saw none value")
+        return value
+    except IndexError:
+        return default
+
+
 def _get_normalized_location(site: dict, timestamp: str) -> schema.NormalizedLocation:
     return schema.NormalizedLocation(
         id=_get_id(site),
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -23,7 +23,9 @@ def _get_availability(site: dict) -> schema.Availability:
 
     appt_options = {
         "Yes": True,
-        "No": False
+        "No": False,
+        "Vax only": True,
+        "Test only": False,
     }
 
     avail = try_lookup(appt_options, appt_only, None, name="availability lookup")
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -195,10 +195,6 @@ def _get_normalized_location(site: dict, timestamp: str) -> schema.NormalizedLoc
     state = site["attributes"]["State"] or None
     state = state.strip() if state is not None else None
 
-    if state not in ['AL', 'AK', 'AS', 'AZ', 'AR', 'CA', 'CO', 'CT', 'DE', 'DC', 'FL', 'GA', 'GU', 'HI', 'ID', 'IL', 'IN', 'IA', 'KS', 'KY', 'LA', 'ME', 'MD', 'MA', 'MI', 'MN', 'MS', 'MO', 'MT', 'NE', 'NV', 'NH', 'NJ', 'NM', 'NY', 'NC', 'ND', 'MP', 'OH', 'OK', 'OR', 'PA', 'PR', 'RI', 'SC', 'SD', 'TN', 'TX', 'UT', 'VT', 'VA', 'VI', 'WA', 'WV', 'WI', 'WY']:
-        print(type(state) or "None")
-        print("\"" + state + "\"" if state is not None else "None")
-
     return schema.NormalizedLocation(
         id=f"{SOURCE_NAME}:{_get_id(site)}",
         name=site["attributes"]["name"],
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -177,7 +177,7 @@ def try_get_lat_long(site):
         )
     except KeyError:
         pass
-    
+
     return location
 
 
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -197,9 +197,9 @@ def _get_normalized_location(site: dict, timestamp: str) -> schema.NormalizedLoc
             zip=try_get_list(city_state_zip, -1, default=""),
         ),
         location=schema.LatLng(
-            latitude=site["geometry"]["y"],
-            longitude=site["geometry"]["x"],
-        ),
+            latitude=site["geometry"]["y"] if site.get("geometry") else None,
+            longitude=site["geometry"]["x"] if site.get("geometry") else None,
+        ) if site.get("geometry") else None,# this is convuluted. ask Adrian (the author) if theres questions
         contact=_get_contacts(site),
         languages=None,
         opening_dates=None,
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -15,6 +15,9 @@ from vaccine_feed_ingest.utils.log import getLogger
 logger = getLogger(__file__)
 
 
+SOURCE_NAME = "us_giscorps_vaccine_providers"
+
+
 def _get_availability(site: dict) -> schema.Availability:
     appt_only = site["attributes"]["appt_only"]
     if appt_only == "Yes":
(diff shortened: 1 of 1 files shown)


MoralCode:
i should add that finding a way to link source URLs that use the automated arcGIS fetcher (like the line `fetched_from_uri` when setting the source) may be another thing that needs to be done as its not particularly easy to get to the link when all you have is the serviceItemId from the fetcher

MoralCode:
once/if #267 is merged this should be ready to go apart from the operating hours field which still isn't done. is this important enough/necessary to have before merging?

MoralCode:
This should now be good to go. I wouldn't say it's 100% complete (most notably the opening hours still need to be parsed from the arbitrary strings. might just want to throw those into the notes for now). but it should at least work well enough to get the data points into the list.

eliblock:
@MoralCode - when I run this locally I get `ModuleNotFoundError: No module named 'schema'`. It looks like this might not be importing the schema in the same way as other normalizers?

I'm going to push a couple of commits to update to the new logger and make sure the schema is included, then can take another look.

Note for reviewers: `us/*` is disabled in the `Run all-stages for modified runners` CI check, so this PR needs to be checked locally.

eliblock (changes requested):
I rebased this + updated to the schema announced in #278, and pushed results here.

When I run this locally, I see several validation errors (for example, the `source` + `id` didn't pass validation - I fixed that as a sample).

Could you take a look at ensuring the remaining data passes validation, and then we can land this code? It is totally okay to add something basic to start and then improve it over time by parsing more data.

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -0,0 +1,264 @@
...
+
+SOURCE_NAME = "us_giscorps_vaccine_providers"
+
+
+def _get_availability(site: dict) -> schema.Availability:
+    appt_only = site["attributes"]["appt_only"]
+    if appt_only == "Yes":
+        return schema.Availability(appointments=True)
+    elif appt_only is not None:
+        return schema.Availability(drop_in=True)
```
eliblock, line 26: Are we confident that "walk-in" should be the default value for a location that is not `appt_only` (vs. unknown)

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -0,0 +1,264 @@
...
+    # inventory = []
+
+    # for vf in vaccines_field:
+    #     try:
+    #         inventory.append(potentials[vf])
+    #     except KeyError as e:
+    #         logger.error("Unexpected vaccine type: %s", e)
+
+    # if len(inventory) > 0:
+    #     return inventory
```
eliblock, line 61: Do we plan to capture this in the future? If not, let's remove the method!

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -0,0 +1,264 @@
...
+    if len(contacts) > 0:
+        return contacts
+
+    return None
+
+
+def sanitize_url(url):
+    url = url.strip()
+    url = url.replace("#", "")
+    url = url.replace("\\", "/")  # thanks windows
```
eliblock, line 97: 😩

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -0,0 +1,264 @@
...
+
+    addrsplit = site["attributes"]["fulladdr"].split(", ")
+
+    zip = site["attributes"]["fulladdr"][-5:]
+    zip = zip if zip.isnumeric() else None
+
+    city_state_zip = addrsplit[1].split(" ") if try_get_list(addrsplit, 1) else None
+
+    state = site["attributes"]["State"] or None
+    state = state.strip() if state is not None else None
```
eliblock, line 196: This looks good for potential follow-up!

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -0,0 +1,264 @@
...
+            street2=None,
+            city=site["attributes"]["municipality"]
+            or try_get_list(city_state_zip, -3, default=""),
+            state=state,
+            zip=zip,
+        ),
+        location=schema.LatLng(
+            latitude=site["geometry"]["y"] if site.get("geometry") else None,
+            longitude=site["geometry"]["x"] if site.get("geometry") else None,
+        )
```
eliblock, line 212: `latitude` and `longitude` are both required if a `LatLng` is set - do we need the `if`s in here?

I see the note below to ask about how it's convoluted 🙃.

Since this is scraped from an ArcGIS source, I'm also not too concerned about these being null, so going to ignore this before merging.

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -0,0 +1,264 @@
...
+        inventory=_get_inventory(site),
+        access=_get_access(site),
+        parent_organization=None,
+        links=None,  # TODO
+        notes=_get_notes(site),
+        active=_get_active(site),
+        source=schema.Source(
+            source=SOURCE_NAME,
+            id=site["attributes"]["GlobalID"],
+            fetched_from_uri="https://services1.arcgis.com/WzFsmainVTuD5KML/ArcGIS/rest/services/COVID19_Vaccine_Site_Survey_API/FeatureServer/0",  # noqa: E501
```
eliblock, line 229: 🤔 this looks like the `ak/arcgis` url.

Where was this data fetched from?

eliblock (changes requested):
@MoralCode - I'm supportive of landing this normalizer and then continuing to improve over time!

One thing that we need to fix before merging is the `fetched_from_uri`. Could you up date that to the proper link... and then we'll be good to go?

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: i dont believe thsi dataset ever had the intention of including this data. will remove

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: yeah, this isnt the best but there were a couple URLS in the dataset that were causing problems (as i mentioned in the discord)

one is literally a physical real-world address (that gets caught by the check for spaces in the url below this), one ended in a # and for some reason that broke the validation, and then there was the one with backslashes in it

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: yep, this is more hacky than anything else. might be ripe to find an address-parsing library similar to my library for opening hours and use that instead

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: yeah, essentially the line after it is supposed to basically make the whole thing null if the "geometry" key is not set, but looking at this again, a try/catch may be better here

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: this data was fetched from the GIScorps dataset. i didnt change this URL because all thats listed in the fetch/parse stages is the serviceItemId and i was feeling lazy and didnt want to try and get the link back out from that. I'll go find the link and swap it out

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: was not aware unknown was an option. will update

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
MoralCode: https://github.com/datamade/usaddress looks good for this. should this wait for another PR to make this review easier?

eliblock: I'm going to merge this - let's add address improvements in a followup!

eliblock (approved):
Thanks - let's get this data moving (and can improve it later! Please open followups!)

PR #147, opened 2021-04-24
Add Colorado fetcher

mbacchi:


MoralCode:
it might be pretty straightforward to use something like beautifulsoup (for python)with maybe some manual parsing to get to just the data within the fetcher so that the parser isn't any more complicated

eliblock (changes requested):
Regarding the above discussion - let's stick with directly fetching the data in the fetcher, and parsing in the parser (unless it's _easier_ to use beautifulsoup to handle the fetching!)

That said, I agree with Raj: could we fetch the map specifically?

eliblock (approved):
😬 This one is going to be fun to `parse`.

Thank you for making the updates!

Issue #155, opened 2021-04-24
fetch vaccinate_gov from Oklahoma


Vaccinebot:
[![learn our pipeline: fetch](https://img.shields.io/static/v1?label=learn%20our%20pipeline&message=fetch&style=social)](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/Runner-pipeline-stages#fetch)

Fetch data from: https://vaccinate.oklahoma.gov/en-US/vaccine-centers/

Put your script in a file named: `ok/vaccinate_gov/fetch.sh` (or `fetch.py`, etc.)

Store it (without processing) in a new file created in the directory passed as the first argument (`sys.argv[1]`).

Check the wiki to learn more about the purpose of the fetch stage and how to get set up for development!

### Tips

1. While working on your code, run it at any point:
    ```sh
    poetry run vaccine-feed-ingest fetch ok/vaccinate_gov
    ```

cdjj:
I can take this.

tmb:
👋 i'll take this.

Sebmaster:
@tmb Just saw this as well. I've actually just finished retrieving the locations. If you're already PRing it, it's all yours though.

tmb:
@Sebmaster ah, shoot. yeah am PRing now — apologies

eliblock:
Sorry about the conflict there - but thank you both for your work today!

tmb:
Fixed by #166

PR #166, opened 2021-04-25
Add Oklahoma vaccinategov fetcher/parser/normalizer

tmb:
filling #155, #156, #157

I also added BeautifulSoup to the poetry deps for scraping — I'm not totally sure if that makes sense to do in this PR so I can move it out / swap to an alternative scraping library if needed.


tmb:
getting caught up on the normalizer here. the vaccinategov site only returns lat/long — is there some built in tool to parse that to an address?

vaccine_feed_ingest/runners/ok/vaccinate_gov/parse.py
```diff
@@ -0,0 +1,24 @@
...
+import json
+import pathlib
+import sys
+
+output_dir = pathlib.Path(sys.argv[1])
+input_dir = pathlib.Path(sys.argv[2])
+
+json_filepaths = input_dir.glob("*.json")
+
+for in_filepath in json_filepaths:
```
Sebmaster, line 14: I think you can replace this whole file with the [built-in infra for ndjson](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/JSON-to-NDJSON-Parser) since it doesn't require JSON traversal (unlike the CA file). Something like:

```
---
state: ok
site: vaccinate_gov
parser: json_list
```

eliblock (changes requested):
@tmb thank you for writing this! This is a great start. Three suggestions:

1. I think @Sebmaster is right - this is well suited for the [shared `json` to `ndjson` parser](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/JSON-to-NDJSON-Parser). Would you mind switching over to using that?
1. Don't worry about getting the `normalize` stage done in this PR... let's land what you have and you can follow up with `normalize`!
1. If the source data does not have addresses, you do not need to include them in the normalized data. Most fields on that schema are optional because we want to accept whatever data we can get! We have another process which matches different datasets and "hydrates" them with lookups where possible. `lat`/`long` is great!

eliblock:
Also - you can fix the current lint error by running `poetry run isort vaccine_feed_ingest/runners/ok/vaccinate_gov`!

vaccine_feed_ingest/runners/ok/vaccinate_gov/parse.py
tmb: thanks! fixed :)

tmb:
@eliblock thanks! should be all set now.

tmb:
ended up getting normalize done so I've tacked that on.

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
```diff
@@ -0,0 +1,117 @@
...
+def _get_address(site: dict) -> schema.Address:
+    CITY_RE = re.compile(r"^(.+?),")
+    ZIP_RE = re.compile(r"\s+(\d{5})")
+
+    raw_address = site["Description"]
+    sections = raw_address.split("\r\n")
+
+    # sometimes they just use \n instead of \r\n
+    if len(sections) == 1:
+        sections = raw_address.split("\n")
```
eliblock, line 32: 🙃 thanks for handling this case.

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def normalize(site: dict, timestamp: str) -> str:
+    """
+    sample:
+    {"Description": "608 NW 9th St\r\nSuite 1100\r\nOklahoma City, Oklahoma 73102 <br>Phone No: 405-425-4489 <br> ", "Distance": 0.6, "Id": "6953713d-7e6a-eb11-a812-001dd800ac9e", "Latitude": 35.47689, "Location": "35.47689,-97.52351", "Longitude": -97.52351, "PushpinImageHeight": 39, "PushpinImageUrl": "", "PushpinImageWidth": 32, "Title": "1st Dose- Pfizer- OKC- SSM Health Family Medicine Center", "Url": ""} # noqa: E501
+    """
+    normalized = schema.NormalizedLocation(
+        id=f"ok_vaccinate_gov:{site['Id']}",
+        name=site["Title"],
```
eliblock, line 95: It looks like this field contains _semi_-structured names _and_ vaccine information.

Do you think it would be possible to pull out the `Name` without the `1st dose Pfizer` part for this field?

Ideally we could also grab the vaccine information to fill out the relevant portion of the schema.

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
```diff
@@ -0,0 +1,117 @@
...
+input_dir = pathlib.Path(sys.argv[2])
+input_file = input_dir / "output.parsed.ndjson"
+output_dir = pathlib.Path(sys.argv[1])
+output_file = output_dir / "output.normalized.ndjson"
+
+with input_file.open() as parsed_lines:
+    with output_file.open("w") as fout:
+        for line in parsed_lines:
+            site = json.loads(line)
+            normalized_site = normalize(site, parsed_at_timestamp)
```
eliblock, line 115: It looks to me like the input data contains some duplicates - e.g., `Integris - Oklahoma Christian University` appears to have multiple entries at the same location but with different vaccine types.

Do you see an easy way to de-dupe these entries? CC: @rajbot who might have some thoughts.

eliblock (changes requested):
Very excited to bring this dataset online! I'm hoping we can parse the name field down a bit further, but other than that... this looks good to go!

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
tmb: lol — yup :)

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
tmb: hm — I can try. some of the names only have "1st dose" or "2nd dose" without a vaccine type, or solely the vaccine type.

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
tmb: Confirmed in Discord that de-duping shouldn't happen in the normalizer stage.

vaccine_feed_ingest/runners/ok/vaccinate_gov/normalize.py
```diff
@@ -0,0 +1,181 @@
...
+    johnson = re.search("janssen", name, re.IGNORECASE) or re.search(
+        "johnson", name, re.IGNORECASE
+    )
+
+    if pfizer:
+        inventory.append(schema.Vaccine(vaccine="pfizer_biontech"))
+    elif moderna:
+        inventory.append(schema.Vaccine(vaccine="moderna"))
+    elif johnson:
+        inventory.append(schema.Vaccine(vaccine="johnson_johnson_janssen"))
```
eliblock, line 105: Great approach!

eliblock (approved):

PR #173, opened 2021-04-25
fix: make setup script executable

eliblock:


eliblock:
FYI @obra re: executable setup.sh

PR #193, opened 2021-04-25
Fetch & parse GA DPH site

Sebmaster:
I've tried keeping the HTML parsing code pretty flexible to keep most of the actual logic in the normalization stage - I guess that'll somewhat flow into each other as we start processing more HTML.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -124,6 +124,21 @@ optional = false
 python-versions = "*"
 
 [[package]]
+name = "beautifulsoup4"
+version = "4.9.3"
+description = "Screen-scraping library"
+category = "main"
+optional = false
+python-versions = "*"
+
+[package.dependencies]
+soupsieve = {version = ">1.2", markers = "python_version >= \"3.0\""}
+
+[package.extras]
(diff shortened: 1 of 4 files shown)


Graham42:
Pretty much all the fetch/* tasks are taken, so I'm jumping in here to hopefully help by adding a code review 🙂 

As it is this PR looks good to me. 👍 

For a further iteration, it looks like we can parse more specific address information out of each row.

Sample address table cell:
```html
<td
  headers="view-field-address-table-column"
  class="views-field views-field-field-address"
>
  <p class="address" translate="no">
    <span class="address-line1">6700 Douglas Blvd</span><br />
    <span class="locality">Douglasville</span>,
    <span class="administrative-area">GA</span>
    <span class="postal-code">30135</span><br />
    <span class="country"></span>
  </p>
</td>
```

Once we get to normalization, for a [minimal record](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/Normalized-Location-Schema#minimal-record) we're going to want these fields parse out.

PR #196, opened 2021-04-25
Add path support to the shared json_list parser

yotann:
Allows the parser to work even when the location list is not top-level JSON. For example, if the JSON file contains `[{"locations":[...actual list of locations...]}]`, the following parse.yml file would work:

```yaml
---
state: ...
site: ...
parser: json_list
path: [0, "locations"]
```

This will be useful for the il/juvare parser. If this is merged, the [wiki page](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/JSON-to-NDJSON-Parser) should also be updated.


eliblock (approved):
Thank you! I've updated the wiki page you referenced - please take a look.

yotann:
Wiki page looks good!

PR #200, opened 2021-04-25
Validate stage outputs with the option to turn off

bryanculbertson:
Validate the outputs of parse and normalize stages to ensure they match the spec.

Does not upload them if they failed.

This can be turned off with `--no-validate` or `ENABLE_VALIDATE=false`

### DIFF
--- a/vaccine_feed_ingest/stages/ingest.py
+++ b/vaccine_feed_ingest/stages/ingest.py
@@ -10,7 +10,7 @@ import pydantic
 from vaccine_feed_ingest.schema import schema
 
 from . import outputs, site
-from .common import RUNNERS_DIR, PipelineStage, STAGE_OUTPUT_SUFFIX
+from .common import RUNNERS_DIR, STAGE_OUTPUT_SUFFIX, PipelineStage
 
 logger = logging.getLogger("ingest")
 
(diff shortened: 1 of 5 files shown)


vaccine_feed_ingest/stages/ingest.py
```diff
@@ -130,12 +135,23 @@ def run_parse(
             check=True,
         )
 
-        if not outputs.data_exists(parse_output_dir):
+        if not outputs.data_exists(
+            parse_output_dir, suffix=STAGE_OUTPUT_SUFFIX[PipelineStage.PARSE]
+        ):
             logger.warning(
                 "%s for %s returned no data files.", parse_path.name, site_dir.name
```
eliblock, line 142: Could we add the expected suffix to this error message (to help a developer troubleshoot)?
```suggestion
                "%s for %s returned no data files with expected extension %s.", parse_path.name, site_dir.name, STAGE_OUTPUT_SUFFIX[PipelineStage.PARSE]
```

eliblock (approved):
Excited to have this validation - quick tests locally look great! One suggestion.

vaccine_feed_ingest/stages/ingest.py
bryanculbertson: Improved error message here are elsewhere to include extension required

PR #218, opened 2021-04-26
Add a tox lint-fix command

bryanculbertson:
Add a tox command that fixes lint errors instead of failing:
```
poetry run tox -e lint-fix
```

This command should be used locally, but not for an automation.


eliblock:
Could you update the PR template to suggest the new command?

eliblock (approved):
The first-run is quite slow (I guess because we have many dependencies!), but after that... this is neat!

Made a suggestion in the above comment to update the PR template. Other than that... :rocket:

bryanculbertson:
Adjusted the PR template

PR #219, opened 2021-04-26
Issue #210: fix fetch and parse computation for arcgis sites

### LINKED ISSUE #210: `available-sites` command does not properly handle .yml config files
Running the CLI `available-sites` command currently returns `no-fetch` or `no-parse` for sites configured via yml. It should properly recognize these instead!

shashank025:
# 

| Key Details |
|-|
| Resolves #210 |
State:  |
Site:  |

## Notes

A straightforward fix to how `fetch` and `parse` statuses are computed for arcgis based sources, as seen when running the following command:

```
poetry run vaccine-feed-ingest available-sites
```

## Data sample

Before this fix, the output for arcgis sites looked like this:

```
ak/arcgis no-fetch no-parse normalize
```

(even though Alaska has valid fetch and parse yml files). After this fix, the output looks correct:

```
ak/arcgis fetch parse normalize
```
I also verified that none of the output changes for non-arcgis sites.

## Before Opening a PR
- [x] I tested this using the CLI; I ran: `poetry run vaccine-feed-ingest all-stages ak/arcgis`
- [x] I ran auto-formatting:
  - [ ] `poetry run black vaccine_feed_ingest/runners`
  - [ ] `poetry run isort vaccine_feed_ingest/runners`

### DIFF
--- a/vaccine_feed_ingest/cli.py
+++ b/vaccine_feed_ingest/cli.py
@@ -82,8 +82,9 @@ def available_sites(state: Optional[str]) -> None:
     """Print list of available sites, optionally filtered by state"""
 
     for site_dir in site.get_site_dirs_for_state(state):
-        has_fetch = bool(site.find_executeable(site_dir, common.PipelineStage.FETCH))
-        has_parse = bool(site.find_executeable(site_dir, common.PipelineStage.PARSE))
+        is_arcgis_site = site_dir.name.endswith('arcgis')
+        has_fetch = _compute_has_fetch(site_dir, is_arcgis_site)
+        has_parse = _compute_has_parse(site_dir, is_arcgis_site)
         has_normalize = bool(
             site.find_executeable(site_dir, common.PipelineStage.NORMALIZE)
         )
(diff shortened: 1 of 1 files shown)


vaccine_feed_ingest/cli.py
```diff
@@ -82,8 +82,9 @@ def available_sites(state: Optional[str]) -> None:
     """Print list of available sites, optionally filtered by state"""
 
     for site_dir in site.get_site_dirs_for_state(state):
-        has_fetch = bool(site.find_executeable(site_dir, common.PipelineStage.FETCH))
-        has_parse = bool(site.find_executeable(site_dir, common.PipelineStage.PARSE))
+        is_arcgis_site = site_dir.name.endswith('arcgis')
```
eliblock, line 85: Some non-ArcGIS sites, [like `ok/vaccinate_gov`](https://github.com/CAVaccineInventory/vaccine-feed-ingest/tree/main/vaccine_feed_ingest/runners/ok/vaccinate_gov), use a `.yml` file as well.

Instead of filtering for only sites which end in `arcgis`, could we capture the broader use case? The [logic for finding files used when running the commands](https://github.com/CAVaccineInventory/vaccine-feed-ingest/blob/d3fb58da986fbda52f01a6750201e2e4febcf791/vaccine_feed_ingest/stages/ingest.py#L91-L94) might provide a good template to follow.

eliblock (changes requested):
Excited to see this command start being useful again soon!

I think we need to make this a bit more general purpose - see the comment on the code!

vaccine_feed_ingest/cli.py
shashank025: I think I captured the broader use case like you suggested, please take a look!

eliblock (approved):
Thank you for updating - this will help folks interact with the tool!

PR #222, opened 2021-04-26
Fix Access schema with wheelchair enum

bryanculbertson:
1. Change wheelchair to be an enum of `yes`, `partial`, `no`
2. Fix `walk` and `drive` to be bool like in the spec.

Change `sf_gov` to produce wheelchair enum values. It is the only one using the schema right now.


eliblock (approved):
Please update the wiki when you merge this!

PR #267, opened 2021-04-28
count locations in guam as valid

MoralCode:
#144 has some data from locations in guam that was failing the validation. this adds and checks an additional bounding box for locations in guam and (part of) the marina islands.

This PR also makes it easy to add more bounding boxes in future if more places are discovered.

before:

```
04/28/2021 09:24:09 WARNING:ingest:Invalid latitude or longitude in /tmp/tmp8ga_r5p5_normalize_us_giscorps_vaccine_providers/output/c50a1a352e944a66aed98e61952051ef_0.parsed.normalized.ndjson at line 972: latitude=13.471961848187085 longitude=144.740660713503 is outside approved bounds (latitude=MinMax(minimum=-14.549, maximum=71.367) longitude=MinMax(minimum=-179.779, maximum=0.0))
04/28/2021 09:24:09 WARNING:ingest:normalize.py for giscorps_vaccine_providers returned invalid source location ndjson files.
```

after: No warnings and the processing continues

## Before Opening a PR
- [ X] I tested this using the CLI (e.g., `poetry run vaccine-feed-ingest normalize us/giscorps_vaccine_providers`)
- [X ] I ran auto-formatting: `poetry run tox -e lint-fix`


vaccine_feed_ingest/stages/ingest.py
```diff
@@ -324,17 +324,49 @@ def _validate_normalized(output_dir: pathlib.Path) -> bool:
...
+        ) or not boundingbox.longitude.contains(location.longitude):
+            if method == "any":
+                # fail if any single bounding box fails
+                return False, boundingbox
+            else:
+                results.append(False)
+
+        results.append(True)
+
+    if method == "all":
```
bryanculbertson, line 360: This extra flexibility of methods is cool, but also unnecessary. Since we don't need the flexibility I don't want to add code to maintain that we don't need until we need it.

Can you remove `method` and just check to make sure it is in one of the approved bounding boxes, and if it is then return True.

vaccine_feed_ingest/stages/ingest.py
```diff
@@ -324,17 +324,49 @@ def _validate_normalized(output_dir: pathlib.Path) -> bool:
                     return False
 
                 if normalized_location.location:
-                    if not BOUNDING_BOX.latitude.contains(
-                        normalized_location.location.latitude
-                    ) or not BOUNDING_BOX.longitude.contains(
-                        normalized_location.location.longitude
-                    ):
+                    result, failed_bound = validate_bounding_boxes(
+                        normalized_location.location, [BOUNDING_BOX, BOUNDING_BOX_GUAM]
```
bryanculbertson, line 328: Can you either change the method to `validate_approved_bounds` and not have the caller pass in a list of bounds, or add a constant called `APPROVED_BOUNDS` which is a list of bounding boxes.

I want to make it easy for the the caller to do the right thing, and right now this flexibility means future callers might not pass all of the bounds.

bryanculbertson (changes requested):
Thank you for tackling this! I have a request to simplify the API to remove options that we don't plan on using.

My rationale is less code to maintain, and less ways for a caller to do the wrong thing.

It would also be nice to have a test for this now since we have a testing framework! But no need to do that if you don't want to tackle that part. If you do try this then add a test a new file called `test_validation.py` in https://github.com/CAVaccineInventory/vaccine-feed-ingest/tree/main/tests/utils

Thank you!

vaccine_feed_ingest/stages/ingest.py
```diff
@@ -324,17 +326,40 @@ def _validate_normalized(output_dir: pathlib.Path) -> bool:
...
                         )
                         return False
+
+    return True
+
+
+def validate_bounding_boxes(location, bounding_boxes):
+    results = []
+
+    for boundingbox in bounding_boxes:
```
bryanculbertson, line 350: You can simplify this even more if you swap the boolean conditional and return True the first time you met it.

```
for bb in bbs:
  if bb.lat.contains() and bb.long.contains():
    return True

return False

eliblock (approved):
Thank you for adding this! I refactored a bit to move this into the `validation` package (so it can be more easily re-used), and added a test.

PR #268, opened 2021-04-28
chore(deps): bump vaccine_feed_ingest_schema to v0.2.1

eliblock:
Update `vaccine_feed_ingest_schema` to `v0.2.1` ([release notes](https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema/releases/tag/0.2.1)).

After the update, migrate all users of the existing `schema` shared import to import either `load` or `location` (per the release notes).

Finally, migrate additional users of the deprecated in-source `schema` to the published one.

### DIFF
--- a/vaccine_feed_ingest/stages/load.py
+++ b/vaccine_feed_ingest/stages/load.py
@@ -8,7 +8,7 @@ import rtree
 import shapely.geometry
 import urllib3
 import us
-from vaccine_feed_ingest_schema import schema
+from vaccine_feed_ingest_schema import load
 
 from .. import vial
 from ..utils.match import canonicalize_address, get_full_address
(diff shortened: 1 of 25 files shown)


vaccine_feed_ingest/stages/load.py
```diff
@@ -8,7 +8,7 @@
 import shapely.geometry
 import urllib3
 import us
-from vaccine_feed_ingest_schema import schema
```
bryanculbertson, line 11: you also need to import location and use that

vaccine_feed_ingest/stages/load.py
```diff
@@ -58,7 +58,7 @@ def run_load_to_vial(
         with filepath.open() as src_file:
             for line in src_file:
                 try:
-                    normalized_location = schema.NormalizedLocation.parse_raw(line)
+                    normalized_location = load.NormalizedLocation.parse_raw(line)
```
bryanculbertson, line 61: this should be location.NormalizeLocation

bryanculbertson (changes requested):
Mostly good, just need to fix the places where are you importing NormalizedLocation indirectly instead of directly from `location`

bryanculbertson (approved):

Issue #233, opened 2021-04-27
convert outstanding existing normalize stages to use typed helpers


eliblock:
[![learn our pipeline: normalize](https://img.shields.io/static/v1?label=learn%20our%20pipeline&message=normalize&style=social)](https://github.com/CAVaccineInventory/vaccine-feed-ingest/wiki/Runner-pipeline-stages#normalize)

Some of the earliest-created normalizers construct the output `.ndjson` directly (rather than using the [packaged schema](https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema)). This means that they do not benefit from type-checking, will not benefit from improvements we plan to make to the package (e.g., auto-`.strip()`-ing strings), and generally make the pipeline more fragile as we iterate on shared components.

Update the following normalizers to use the [published/packaged schema helpers](https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema):
- [x] `ca/sf_gov`
- [x] `ct/covidvaccinefinder_gov`
- [x] `ma/vaxfinder_gov` (note: this site will also be updated in #223)
- [x] `ny/am_i_eligible_covid19vaccine_gov` (note: mostly uses the schema helpers, but appends partial addresses directly)
- [x] `us/vaccinespotter_org`

<hr />

To import the schema helpers:

```python
from vaccine_feed_ingest_schema import schema
```

<hr />

### Tips

1. Fetch and parse data for this site before you start developing:
    ```sh
    poetry run vaccine-feed-ingest fetch <state>/<site>
    poetry run vaccine-feed-ingest parse <state>/<site>
    ```

1. While working on your code, run it at any point:
    ```sh
    poetry run vaccine-feed-ingest normalize <state>/<site>
    ```

shaqq:
currently tackling ca/sf_gov right now

shaqq:
PR out!

eliblock:
Thank you!

FYI, we updated the schema for address earlier today to make all fields optional. If you re-install the poetry dependencies, you should pick that up! I'll update the wiki now.

shaqq:
👍 just updated my PR

#276 _should_ cover the rest of this then, but would be good to have someone double-check

eliblock:
It looks to me like they have all been converted!

As I mentioned in my review of #276, it looks like there are a few places we will need to continue tweaking to fully take advantage of the schema, but per #278 it seems like it makes sense to wait for the validations to be in place before taking another look.

[...]

PR #276, opened 2021-04-29
Use schema in earlier normalizers

shaqq:
| Key Details |
|-|
| Related #233 |

## Notes
Used the Pydantic schema for the following:

- ca/sf_gov
- ct/covidvaccinefinder_gov
- us/vaccinespotter_org
- ny/am_i_eligible_covid19vaccine_gov

Additionally, when updating `us/vaccinespotter_org`, we could be using `site["provider_brand_name"]` and `site["provider_brand_id"]` to populate the links & parent_org. There were a few edge cases I didn't want to tackle, but figured I should call it out here.

## Before Opening a PR
- [x] I tested this using the CLI (e.g., `poetry run vaccine-feed-ingest <state>/<site>`)
- [x] I ran auto-formatting: `poetry run tox -e lint-fix`

### DIFF
--- a/vaccine_feed_ingest/runners/ca/sf_gov/normalize.py
+++ b/vaccine_feed_ingest/runners/ca/sf_gov/normalize.py
@@ -5,8 +5,9 @@ import json
 import pathlib
 import sys
 
-from vaccine_feed_ingest.utils.normalize import provider_id_from_name
 from vaccine_feed_ingest_schema import location as schema
+
+from vaccine_feed_ingest.utils.normalize import provider_id_from_name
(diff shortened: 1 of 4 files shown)


eliblock (approved):
Thank you for doing this conversion!

In poking through these, I think there's a few places we will be able to tweak further to get closer to the schema, but per #278 it seems like it makes sense to wait for further changes until the validation is in place (ideally tomorrow).

I'll leave the issue open since you mentioned a potential followup in your PR, but feel free to close it if you think we're at a stopping point!

Issue #278, opened 2021-04-29
[Announcement] New required validation in Normalizers


Vaccinebot:
## Thank You

Thank you to everyone who has contributed to this project so far. Your efforts have already begun to build out a comprehensive dataset of the available vaccination locations in the US.

## An issue we're facing

Over the past several days we have been working on downstream pipelines that take this scraped data and help it make its way to our data sharing partners and [vaccinateTheStates.com](https://www.vaccinateTheStates.com).

One thing that has become clear is that we need to normalize data a bit more precisely to be able to consistently handle it in downstream systems.

## What's we did

Version `v1.0.0` of the schema includes additional validation on several fields in the schema. The full details of the new validations may be found [in the release notes](https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema/releases/tag/1.0.0) - please read them to find out what changed.

Introducing these validations has broken most (if not all) existing normalizers. We will be opening issues to fix them.

Additionally, we added a few required PR checks:
* One check enforces that the schema upgrade is included in PRs
* One check fixes existing `isort` issues with SuperLinter, and runs basic tests using `pytest`
* One check runs any modified `fetch`/`parse`/`normalize` stages (exceptions `us/*` are not run).

We know these changes are disruptive. Once the validation issues are fixed, however, the data coming out of the pipeline will be much more usable by downstream systems, including data sharing partners and [vaccinateTheStates.com](https://www.vaccinateTheStates.com).

## What now?

If you have an open PR, please rebase it on `main`, run `poetry install --extras lint` again, and ensure it complies with the validations.

If you are looking to get involved, please pick up an issue to fix an existing normalizer.

Once these changes are complete, we plan to begin introducing issues to `fetch`/`parse`/`normalize` additional datasets (many of them county-level or territory-level rather than state-level) which will help surface locations not widely advertised elsewhere. You have seen some examples of this already (e.g., #183 and #189), and we are excited to add many more!

PR #287, opened 2021-04-29
ci: force rebase job

eliblock:
As shared in #278, we plan to release `v1.0.0` of the schema. When it is released, all open PRs will need to be updated to include it properly.

Once the PR bumping the schema is merged, merge this PR and require it as a PR check.
* Old branches will fail the required check (because it is not defined)
  image
* New branches will pass the required check (because it simply echos a value)
  image


bryanculbertson (approved):
This is a clever way to force a rebase. I like it.

Issue #290, opened 2021-04-29
Authorities should not have colons in them


simonw:
One example is this code here: https://github.com/CAVaccineInventory/vaccine-feed-ingest/blob/90a242bbe1814f54d63aa84a1b0949f60a88cf4a/vaccine_feed_ingest/runners/ct/covidvaccinefinder_gov/normalize.py#L55-L59

VIAL expects that authorities do not include colons - so if it sees `foo:123:2342354:2341` it will treat that as authority = `foo`, identifier = `123:2342354:2341`

Discord: https://discord.com/channels/799147121357881364/824443781990187008/837397802501275699

simonw:
Here are the authorities we've seen so far in VIAL with this problem:
```
ak:arcgis
az:arcgis
ct_gov:network_id
in:arcgis
ky:govstatus
me:maine_gov
mo:arcgis
nc:myspot_gov
pa:arcgis
ri:arcgis
sc:arcgis
```
We agreed on Discord [here](https://discord.com/channels/799147121357881364/824443781990187008/837401149950525452) to convert those `:` to underscores.

simonw:
I think this is fixed. We can reopen if it crops up again.

simonw:
That migration has now run on staging: https://vial-staging.calltheshots.us/admin/core/concordanceidentifier/ - and on production: https://vial.calltheshots.us/admin/core/concordanceidentifier/

Discord progress here: https://discord.com/channels/799147121357881364/824443781990187008/837903902988566548

simonw:
Here's the before and after for production:

Before/ After:


<img width="273" alt="Select_concordance_identifier_to_change___VIAL_admin" src="https://user-images.githubusercontent.com/9599/116770531-792ee780-a9f9-11eb-82d8-8d9b88c9056c.png"> <img width="273" alt="Select_concordance_identifier_to_change___VIAL_admin" src="https://user-images.githubusercontent.com/9599/116770527-73d19d00-a9f9-11eb-8888-8ee1011fc56b.png">

PR #291, opened 2021-04-29
Use underscores rather than colons in authorities

simonw:
Refn #290

### DIFF
--- a/vaccine_feed_ingest/runners/ak/arcgis/normalize.py
+++ b/vaccine_feed_ingest/runners/ak/arcgis/normalize.py
@@ -49,7 +49,7 @@ def _get_id(site: dict) -> str:
     arcgis = "d92cbd6ff2524d7e92bef109f30cb366"
     layer = 0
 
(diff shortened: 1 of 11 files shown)


bryanculbertson:
@simonw You also need to check where each of them make their ids and fix those too. The id prefix should match the source.

simonw:
Oh I see, places like this: https://github.com/CAVaccineInventory/vaccine-feed-ingest/blob/90a242bbe1814f54d63aa84a1b0949f60a88cf4a/vaccine_feed_ingest/runners/az/arcgis/normalize.py#L33-L46

simonw:
I'm testing this by running:

    poetry run vaccine-feed-ingest all-stages

Then I'm going to check all of the generated normalized output to see what their authority values look like.

simonw:
I explored the output a bit in a Jupyter notebook here: https://gist.github.com/simonw/cfa7acf9d9e7ac10696f978016704ef1 - it looks like this does the right thing.

bryanculbertson (approved):
Thank you!

bryanculbertson:
@simonw I am going to merge this

PR #285, opened 2021-04-29
Refactor stage

shashank025:
# 

| Key Details |
|-|
| Resolves #255 |
State:  |
Site:  |

## Notes

NA

## Data sample

Ran `all-stages` for `ak/arcgis`via:

```
 poetry run vaccine-feed-ingest all-stages ak/arcgis
```

and examined the first line of normalization output via:

```
head -1 out/ak/arcgis/normalized/2021-04-28T22\:44\:45/d92cbd6ff2524d7e92bef109f30cb366_0.parsed.normalized.ndjson
```

```json
[...]

PR #301, opened 2021-04-30
Rebase #285 onto main

bryanculbertson:


bryanculbertson:
Looks like this was already figured out so I am closing

PR #308, opened 2021-04-30
Add enrich stage between normalize and load

bryanculbertson:
Add a bare bones `enrich` stage that processes the normalized records from ingestors and prepares them for load.

```
vaccine-feed-ingest enrich ca/sf_gov
```

Right now this stage copies all locations that pass validation, but in the future it will:
1. Ensure a full address for all locations (street, city, state, zip) if possible
2. Ensure lat-lng exists using a geocoder
3. Fill out parent_organization based on name
4. Fill out concordances from parent_organization, source, and name
5. Match to external concordances like google places, OSM POI, or Placekey (if api pricing and limits allow)

### DIFF
--- a/vaccine_feed_ingest/cli.py
+++ b/vaccine_feed_ingest/cli.py
@@ -200,6 +200,25 @@ def all_stages(
 
 
 @cli.command()
+@_state_option()
+@_output_dir_option()
+@_dry_run_option()
+@_sites_argument()
+def enrich(
+    state: Optional[str],
+    output_dir: pathlib.Path,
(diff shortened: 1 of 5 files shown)


eliblock (approved):
This seems like a reasonable direction to go! When this merges, we'll need to update the pipeline runners to call this stage.

Note: I did not test locally.

bryanculbertson:
@eliblock Yeah, I am improving the way we call multiple steps in a different PR here: https://github.com/CAVaccineInventory/vaccine-feed-ingest/pull/311

PR #331, opened 2021-05-01
Update ingest schema to `1.2`

bryanculbertson:
Update ingest schema to `1.2` and also update libs in lock file to latest with `poetry update`


bryanculbertson:
Eli beat me to it: https://github.com/CAVaccineInventory/vaccine-feed-ingest/pull/330

PR #352, opened 2021-05-02
normalize: fix validation errors in normalize for nc/myspot_gov (#339).

nschiefer:
The phone numbers and websites were extremely messy, so I tried to capture what I could with regular expressions.

## Before Opening a PR
- [x] I tested this using the CLI (e.g., `poetry run vaccine-feed-ingest <state>/<site>`)
- [x] I ran auto-formatting: `poetry run tox -e lint-fix`

### DIFF
--- a/vaccine_feed_ingest/runners/nc/myspot_gov/normalize.py
+++ b/vaccine_feed_ingest/runners/nc/myspot_gov/normalize.py
@@ -3,11 +3,13 @@
 import datetime
 import json
 import pathlib
+import re
 import sys
 
 from vaccine_feed_ingest_schema import location as schema
 
 from vaccine_feed_ingest.utils.log import getLogger
+from vaccine_feed_ingest.utils.normalize import normalize_url
 
 logger = getLogger(__file__)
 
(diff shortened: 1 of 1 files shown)


vaccine_feed_ingest/runners/nc/myspot_gov/normalize.py
```diff
@@ -43,27 +45,42 @@ def _get_location(site: dict):
...
-            phone_notes = phone_notes.lstrip(" ")
-            ret.append(schema.Contact(phone=phone, other=f"phone_notes:{phone_notes}"))
+        url = normalize_url(url)
+        ret.append(schema.Contact(website=url))
+    return ret
+
+
+phone_number_parser = re.compile(
+    r"^(?:(?:\+?1\s*(?:[.-]\s*)?)?(?:\(?([0-9]{3})\)?)\s*(?:[.-]\s*)?)?([0-9]{3})\s*(?:[.-]\s*)?([0-9]{4})\s*(?:\([A-Z]{4}\))?,?(?:\s*(?:\#|x\.?|ext\.?|extension|OPTION|Option|option|Ext)\s*(\d+))?"  # noqa: E501
+)
```
eliblock, line 66: Can we potentially use the regex directly from `vaccine_feed_ingest_schema` by importing it (or is this a _different_ regex)?

vaccine_feed_ingest/runners/nc/myspot_gov/normalize.py
```diff
@@ -43,27 +45,42 @@ def _get_location(site: dict):
...
-            phone = "(" + raw_phone[0:3] + ") " + raw_phone[4:7] + "-" + raw_phone[8:12]
-            phone_notes = raw_phone[12:]
-        elif len(raw_phone) == 10:
-            phone = "(" + raw_phone[0:3] + ") " + raw_phone[3:6] + "-" + raw_phone[6:10]
-            phone_notes = ""
+        ret.extend(_parse_phone_numbers(site["Appointment Phone"]))
+
+    url = site["Web Address"]
+    # Some URLs have multiple schemes.
+    valid_url = re.match(r"(https?:\/\[/)*:(.+)", url)
```
eliblock, line 52: Why is there a `\[` in the regex? Should this be more like:
```suggestion
    valid_url = re.match(r"(https?:\/\/)?(.+)", url)
```

vaccine_feed_ingest/runners/nc/myspot_gov/normalize.py
nschiefer: It's a different regex :(

In particular, it is much more flexible about formatting, since I'm trying to parse rather than validate.

vaccine_feed_ingest/runners/nc/myspot_gov/normalize.py
nschiefer: That's bizarre and I'm glad that you caught it before we threw out a bunch of websites. I'm going to verify that this still works.

nschiefer: And indeed, it stopped working!

eliblock (approved):
Thank you for fixing!

PR #358, opened 2021-05-02
Enrich locations with provider info from name

bryanculbertson:
In the enrichment stage, add the provider from `normalize.provider_id_from_name` as a link if it is missing. It will not overwrite values if they were added in the normalize step.

This means we don't have to add it in all of the normalize stages if we don't want to.

### DIFF
--- a/vaccine_feed_ingest/stages/enrichment.py
+++ b/vaccine_feed_ingest/stages/enrichment.py
@@ -1,11 +1,13 @@
 """Method for enriching location records after that are normalized"""
 import pathlib
+from typing import Dict, Optional
 
 import pydantic
 from vaccine_feed_ingest_schema import location
 
 from vaccine_feed_ingest.utils.log import getLogger
 
+from ..utils import normalize
 from . import outputs
 from .common import STAGE_OUTPUT_SUFFIX, PipelineStage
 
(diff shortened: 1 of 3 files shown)


vaccine_feed_ingest/stages/enrichment.py
```diff
@@ -52,6 +54,45 @@ def enrich_locations(input_dir: pathlib.Path, output_dir: pathlib.Path) -> bool:
...
+
+    provider_authority, provider_id = provider_tuple
+
+    existing_links = _generate_link_map(loc)
+
+    if str(provider_authority) not in existing_links:
+        loc.links = [
+            *(loc.links or []),
+            location.Link(authority=provider_authority, id=provider_id),
+        ]
```
eliblock, line 95: Should this _also_ add an `authority` if the `provider_authority` does exist but is set to a different ID?

I could see arguments for this either way.

eliblock (approved):

vaccine_feed_ingest/stages/enrichment.py
bryanculbertson: I would prefer to play it safe since this is automatically added it.

If the normalizer added a different id then the normalizer may have had a good reason for it, and i want to allow the normalizer to have override privileges here.

PR #359, opened 2021-05-02
Enrich concordance links with a link for source if one is missing

bryanculbertson:
Add concordance link for source in enrichment stage if one is missing

### DIFF
--- a/vaccine_feed_ingest/stages/enrichment.py
+++ b/vaccine_feed_ingest/stages/enrichment.py
@@ -61,6 +61,7 @@ def _process_location(
     enriched_location = normalized_location.copy()
 
     _add_provider_from_name(enriched_location)
+    _add_source_link(enriched_location)
 
     return enriched_location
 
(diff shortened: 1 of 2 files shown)


vaccine_feed_ingest/stages/enrichment.py
```diff
@@ -96,3 +97,22 @@ def _add_provider_from_name(loc: location.NormalizedLocation) -> None:
...
+    if not loc.source:
+        return
+
+    if not loc.source.source or not loc.source.id:
+        return
+
+    existing_links = _generate_link_map(loc)
+
+    if str(loc.source.source) in existing_links:
+        return
```
eliblock, line 113: Similar question to the other PR - should we check that this has the "right" id?

eliblock (approved):

vaccine_feed_ingest/stages/enrichment.py
bryanculbertson: If the normalizer added a different id then i assume they had a good reason at this stage.

If we want to do "validation" then we should do that in the schema.

Issue #254, opened 2021-04-28
Expand `provider_id_name()` to cover more chains


nschiefer:
We have a function (`provider_id_from_name()`) that tries to parse retail pharmacy store numbers from a string with the location's name (e.g., "Walgreens #7150 -> 7150).

We want to support this on every chain of retail pharmacies that we know about.

tiffting:
Who all is missing from [`provider_id_from_name()`](https://github.com/CAVaccineInventory/vaccine-feed-ingest/blob/e72eb49e21bac6da07effc0abdad1af0df19e109/vaccine_feed_ingest/utils/normalize.py#L8)? Is there an [authoritative list](https://en.wikipedia.org/wiki/List_of_pharmacies#United_States) of retail pharmacy chains we're looking to support?

tiffting:
Here's the [CDC's list](https://www.cdc.gov/vaccines/covid-19/retail-pharmacy-program/participating-pharmacies.html) of participating pharmacies.

nschiefer:
The target is to understand as many of the formats that we already have. There are a few ways that we could get this. The easiest _might_ be to the run the fetcher/parser/normalizer phases locally (using the instructions on this repo's wiki) and look for patterns. Alternatively, I could dump the list of places that we've found so far from our database.

nschiefer:
Here is a random sample of 10k location names, which should be pretty representative of what we're hoping to parse. [select-name-from-location-orded0ba3a.csv](https://github.com/CAVaccineInventory/vaccine-feed-ingest/files/6388621/select-name-from-location-orded0ba3a.csv)

shashank025:
I can take this one.

shashank025:
@nschiefer thanks so much for that csv file. I examined it a bit, via:

```
$ sort select-name-from-location-orded0ba3a.csv | uniq | less
```

and I see a bunch of patterns, including some that are _not_ currently handled:

```
CVS Pharmacy, Inc. #00052
CVS Pharmacy, Inc. #00076
...
Community, A Walgreens Pharmacy #13434
Community, A Walgreens Pharmacy #13435
...
Walmart Inc, #1828
Walmart Inc, #2271
Walmart Inc, #2509
...
Acme Pharmacy #1058
Acme Pharmacy #1077
Acme Pharmacy #1092
...
Albertsons Companies Inc. #3311
Albertsons Market Pharmacy #1915
Albertsons Market Pharmacy #3917
[...]

PR #332, opened 2021-05-01
(pt 1) support extraction of additional providers #254

shashank025:
| Key Details |
|-|
| Addresses #254  |
State:  |
Site:  |

## Notes

This PR uses [a sample CSV](https://github.com/CAVaccineInventory/vaccine-feed-ingest/files/6388621/select-name-from-location-orded0ba3a.csv) provided on #254 that contains about 10K rows of parsed provider info to improve coverage of providers while extracting a provider id and name from data sources. Specifically, that CSV was sorted alphabetically, and support was added for providers in that sorted list whose names start from **_A_** through **_K_**.

This is the first in a series of planned steps to improve coverage. Upcoming plans:

1. Add the newly identified providers to the [schema repo](https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema)
2. Support other providers (**_L_** through **_Z_**) from the provided CSV.
3. Add a simple helper script that shows a measure of **_coverage_** by applying this script to a specified input file. Basically, it will spit out how many of the input rows had a matching provider name + id extracted. This helper can be used to track improvements to coverage over time.
4. [Optimization] Rearrange the regexes in `provider_id_from_name()` in the order of expected national coverage, so that a national chain like COSTCO or WALGREENS with heavy coverage can be matched first, and the function can then exit early.

### Update

This PR also updates the `provider_id_from_name()` method to return a `VaccineProvider` (previously `str`).

## Data sample

As a sanity check, I ran:

```
poetry run vaccine-feed-ingest all-stages ok/vaccinate_gov
poetry run vaccine-feed-ingest all-stages ca/sf_gov
```
and examined the normalized outputs.

## Before Opening a PR
- [X] I tested this using the CLI (e.g., `poetry run vaccine-feed-ingest <state>/<site>`)
- [X] I ran auto-formatting: `poetry run tox -e lint-fix`

PR #360, opened 2021-05-02
Support extraction of ids for more providers (L - Z) found in data

shashank025:
| Key Details |
|-|
| Resolves: #254 |
State: |
Site: |

## Notes

Continuation of work started with #332. This PR builds on https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema/pull/21 to add support of extraction of more vaccine providers.

This PR also refactors the provider_id code to make it easier to maintain and extend by extract config out of code.

## Data sample

I will run the runners shortly ...

## Before Opening a PR

### DIFF
--- a/vaccine_feed_ingest/utils/normalize.py
+++ b/vaccine_feed_ingest/utils/normalize.py
@@ -13,7 +13,6 @@ VACCINE_PROVIDER_REGEXES = {
         re.compile(r"ACME PHARMACY #(\d+)", re.I),
     ],
     VaccineProvider.ALBERTSONS: [
-        re.compile(r"SAV-?ON PHARMACY #\s?(\d+)", re.I),
         re.compile(r"ALBERTSONS(?: MARKET)? PHARMACY #(\d+)", re.I),
     ],
     VaccineProvider.BIG_Y: [
@@ -66,9 +65,6 @@ VACCINE_PROVIDER_REGEXES = {
     VaccineProvider.GIANT_FOOD: [
         re.compile(r"GIANT FOOD #(\d+)", re.I),
     ],
-    VaccineProvider.HEB: [
-        re.compile(r"H-E-B #(\d+)", re.I),
-    ],
     VaccineProvider.HAGGEN: [
         re.compile(r"HAGGEN PHARMACY #(\d+)", re.I),
     ],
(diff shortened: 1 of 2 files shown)


vaccine_feed_ingest/utils/normalize.py
```diff
@@ -7,142 +7,118 @@
 import url_normalize
 from vaccine_feed_ingest_schema.location import VaccineProvider
 
+VACCINE_PROVIDER_REGEXES = {
+    VaccineProvider.ALBERTSONS: {
```
bryanculbertson, line 11: I am concerned about this data model. This process combines brands that are run by the same corporation into the same id space. Are you sure that is the way it works in practice?

By this I mean, is Albertsons #5301 and Safeway #5301 the same store, or are they different stores?

bryanculbertson: What do you think of extracting the provider, brand, and id and creating an ID from those parts to ensure uniqueness e.g. `albertsons:osco_5301` or `albertsons_osco:5301`

shashank025: I don't know enough about either our front-end, or how the data is used to truly weigh in, but after reviewing:
https://www.cdc.gov/vaccines/covid-19/retail-pharmacy-program/participating-pharmacies.html - my suggestion is that we enhance the schema to explicitly represent brands, providers, and the relationship between them.

To be more specific:

* The `VaccineProvider` enum hereafter represents individual brands (e.g., Big Y Pharmacy, Brookshires Pharmacy, etc)
* A new enum, say: `VaccineProviderParent` (with values like Topco, Kroger, etc) represents businesses that have one or more `VaccineProvider` subsidiaries
* Add a `getParent()` helper method to  the `VaccineProvider` enum and implement it appropriately to return an `Optional[VaccineProviderParent]`.
* Modify `provider_id_from_name()` to return either a `VaccineProvider` or `VaccineProviderParent` (use the `Union` type).

shashank025: Explicitly representing brands, parents, and their relationship in the schema will make it easier to build and maintain richer frontends showing users these relationships (if necessary).

shashank025: Tagging @nschiefer to weigh in since he created #254

shashank025: Per discussion on Discord, the above proposal is fine, but the adding of the parent enum etc can be done later. For now, this PR will proceed with the assumption that `VaccineProvider` represents individual brands.

shashank025: Done.

shashank025:
The broken tests will be fixed once the repo is updated with https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema/releases/tag/1.2.2

bryanculbertson (approved):
Thank you!

PR #364, opened 2021-05-03
Turn provider regexs into constants and group them by provider. 

bryanculbertson:
Turn provider regexs into constants and group them by provider. 

This was introduced in #360 and I wanted to get that improvement in without while we discuss the merits of umbrella org vs. brand.


shashank025:
This might not work without also getting https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema/pull/21 in.

shashank025:
> This might not work without also getting [CAVaccineInventory/vaccine-feed-ingest-schema#21](https://github.com/CAVaccineInventory/vaccine-feed-ingest-schema/pull/21) in.

I guess it will work because the Python typing is not _real_ :)

PR #544, opened 2021-05-05
chore: remove networkId as link in ct

eliblock:
Linking on networkId is causing locations in the same network (e.g., Walgreens) to match with each other when they should not. Remove the link.


bryanculbertson (approved):
You beat me to it

PR #561, opened 2021-05-07
fix(import_source_locations): Error about missing raise_for_status

shajith:
Followup to #557, to fix the error about `raise_for_status`:

image

### DIFF
--- a/vaccine_feed_ingest/vial.py
+++ b/vaccine_feed_ingest/vial.py
@@ -4,6 +4,7 @@ import contextlib
 import json
 import urllib.parse
 from typing import Any, Iterable, Iterator, Tuple
+from urllib.error import HTTPError
 
 import geojson
 import rtree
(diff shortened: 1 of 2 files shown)


vaccine_feed_ingest/vial.py
```diff
@@ -84,7 +85,14 @@ def import_source_locations(
...
 
-        rsp.raise_for_status()
+        if rsp.status != 200:
+            raise HTTPError(
+                f"/api/importSourceLocations?import_run_id={import_run_id}",
+                rsp.status,
+                rsp.data[:100],
+                dict(rsp.headers),
+                None,
+            )
```
eliblock, line 95: Thanks for setting the full details!

https://github.com/python/cpython/blob/7424105a02bbc209e032752636bae30ccb556402/Lib/urllib/error.py#L39-L44

eliblock (approved):
😅 thank you for fixing this!

PR #562, opened 2021-05-07
Improve matcher to look at phone numbers and providers

bryanculbertson:
1. Move address similarity to helper, and allow for slight differences for a match to count.
2. If providers are available then block matching if providers are dissimilar
3.  If phone numbers are available then block matching if phone numbers are dissimilar

This does not include the promised address matching improvements because I am still figuring out how to make them work on dirty addresses.

### DIFF
--- a/vaccine_feed_ingest/stages/load.py
+++ b/vaccine_feed_ingest/stages/load.py
@@ -259,6 +259,11 @@ def _is_match(source: location.NormalizedLocation, candidate: dict) -> bool:
         # Shared concordances, mark as match
         return True
 
+    # Don't match locations with different providers
+    provider_matches = is_provider_similar(source, candidate, threshold=0.7)
+    if provider_matches is not None and provider_matches is False:
+        return False
+
     address_matches = is_address_similar(source, candidate)
     if address_matches is not None and address_matches is True:
         return True
--- a/vaccine_feed_ingest/stages/load.py
+++ b/vaccine_feed_ingest/stages/load.py
@@ -14,7 +14,9 @@ from vaccine_feed_ingest.utils.log import getLogger
 
 from .. import vial
 from ..utils.match import (
+    has_matching_phone_number,
     is_address_similar,
+    is_provider_similar,
 )
 from . import outputs
 from .common import STAGE_OUTPUT_SUFFIX, PipelineStage
@@ -262,6 +264,11 @@ def _is_match(source: location.NormalizedLocation, candidate: dict) -> bool:
     # Don't match locations with different providers
     provider_matches = is_provider_similar(source, candidate, threshold=0.7)
     if provider_matches is not None and provider_matches is False:
+        return False
+
+    # If there are phone numbers and the phone numbers don't match then fail to match
+    phone_matches = has_matching_phone_number(source, candidate)
+    if phone_matches is not None and phone_matches is False:
         return False
 
     address_matches = is_address_similar(source, candidate)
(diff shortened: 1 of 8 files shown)


vaccine_feed_ingest/utils/match.py
```diff
@@ -1,8 +1,138 @@
...
+    """Calculate a similarity score between providers.
+
+    - True if similarity is above threshold
+    - False if similarity is below threshold
+    - None if couldn't compare providers
+    """
+    if not source.parent_organization:
+        return None
+
+    src_org = source.parent_organization.name or source.parent_organization.id
```
eliblock, line 70: Is this needed because normalizers use these fields inconsistently?

vaccine_feed_ingest/utils/match.py
```diff
@@ -1,8 +1,138 @@
...
+    """
+    candidate_props = candidate.get("properties", {})
+
+    if not candidate_props.get("phone_number"):
+        return None
+
+    try:
+        cand_phone = phonenumbers.parse(candidate_props["phone_number"], "US")
+    except phonenumbers.NumberParseException:
+        logger.warning("Invalid phone number: %s", candidate_props["phone_number"])
```
eliblock, line 113: nit:
```suggestion
        logger.warning("Invalid candidate phone number for location %s: %s", <an_id_field>, candidate_props["phone_number"])
```

vaccine_feed_ingest/utils/match.py
```diff
@@ -1,8 +1,138 @@
...
+
+    src_phones = []
+    for contact in source.contact:
+        if not contact.phone:
+            continue
+
+        try:
+            src_phones.append(phonenumbers.parse(contact.phone, "US"))
+        except phonenumbers.NumberParseException:
+            logger.warning("Invalid phone number: %s", contact.phone)
```
eliblock, line 124: nit:
```suggestion
            logger.warning("Invalid source phone number for location %s: %s", source.id, contact.phone)
```

vaccine_feed_ingest/stages/load.py
```diff
@@ -244,22 +248,32 @@ def _is_match(source: location.NormalizedLocation, candidate: dict) -> bool:
...
-            return True
+    # Don't match locations with different providers
+    provider_matches = is_provider_similar(source, candidate, threshold=0.7)
+    if provider_matches is not None and provider_matches is False:
+        return False
+
+    # If there are phone numbers and the phone numbers don't match then fail to match
+    phone_matches = has_matching_phone_number(source, candidate)
+    if phone_matches is not None and phone_matches is False:
+        return False
```
eliblock, line 272: I wonder how common it is to see candidates which _should_ match but where we have a different phone number than the source.

That said, seems like a good rule at this point!

eliblock (approved):

vaccine_feed_ingest/stages/load.py
bryanculbertson: I am following the Velma channel, and this was inspired by this case of exactly matching address, but different clinics: https://discord.com/channels/799147121357881364/839741667560259604/839748175204188180

vaccine_feed_ingest/utils/match.py
bryanculbertson: Yes. Some of the normalizers just fill out the name.

This is also a little hacky, I would like to match directly to providers, but VIAL doesn't return a strong identifier for their providers in the API yet.

vaccine_feed_ingest/utils/match.py
```diff
@@ -1,8 +1,138 @@
...
+    if not source.address.state or not candidate["properties"].get("state"):
+        return None
+
+    # City must exist to compare
+    if not source.address.city:
+        return None
+
+    # Street must exist to compare
+    if not source.address.street1:
+        return None
```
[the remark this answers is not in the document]
bryanculbertson: Because VIAL is missing them for most of their locations. Address data in VIAL is really dirty.

PR #571, opened 2021-05-07
Skip matching source locations that are already matched

bryanculbertson:
Load source locations from VIAL and skips matching locations that have already been matched.

The source locations will be updated with the latest data just not with any matching information.

### DIFF
--- a/vaccine_feed_ingest/stages/load.py
+++ b/vaccine_feed_ingest/stages/load.py
@@ -32,6 +32,7 @@ def load_sites_to_vial(
     vial_apikey: str,
     enable_match: bool,
     enable_create: bool,
+    enable_rematch: bool,
     match_ids: Optional[Dict[str, str]],
     create_ids: Optional[Collection[str]],
     candidate_distance: float,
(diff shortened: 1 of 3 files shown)


vaccine_feed_ingest/stages/load.py
```diff
@@ -181,13 +201,22 @@ def run_load_to_vial(
...
-        "Imported %d source locations for %s (%d new, %d matched, %d unknown)",
+        "Imported %d source locations for %s "
+        "(%d new, %d matched, %d unknown, %d already matched)",
         num_imported_locations,
         site_dir.name,
         num_new_locations,
         num_match_locations,
-        num_imported_locations - num_new_locations - num_match_locations,
+        num_unknown_locations,
+        num_already_matched_locations,
```
eliblock, line 219: This is really "number of locations which we skipped because they were already matched and we configured the load step to skip matches, otherwise 0."

Could we change the log line slightly to reflect that?

vaccine_feed_ingest/vial.py
```diff
@@ -155,3 +155,41 @@ def update_existing_locations(
...
+    vial_http: urllib3.connectionpool.ConnectionPool,
+) -> Set[str]:
+    """Return all matched source location ids in VIAL"""
+    source_locations = search_source_locations(vial_http, all=1, matched=1)
+
+    return {
+        loc["properties"]["source_uid"]
+        for loc in source_locations
+        if "properties" in loc and "source_uid" in loc["properties"]
+    }
```
eliblock, line 195: 😩 (my thoughts on this styling, though it is powerful!)

eliblock (approved):
Makes sense to me! I think we'll want to update our worker to sometimes run `--match --create --no-rematch` and sometimes run `--match --no-create --rematch` 🤔

bryanculbertson:
@eliblock The default is no `--no-rematch`, but it could make sense to run a re-match every week or so? I am hoping that the identity of the sources don't change often enough to need to re-match.

vaccine_feed_ingest/stages/load.py
bryanculbertson: Updated log line to be conditional on if we are rematching or not

PR #573, opened 2021-05-07
Handle legacy concordances stored as fields on location

bryanculbertson:
Location data in VIAL has legacy concordance fields that have 't fully been copied over to concordances. Use these fields when matching

- google_places_id field -> google_places concordance
- vaccinefinder_location_id -> vaccinefinder_org concordance
- vaccinespotter_location_id -> vaccinespotter_org concordance

### DIFF
--- a/vaccine_feed_ingest/utils/match.py
+++ b/vaccine_feed_ingest/utils/match.py
@@ -13,6 +13,14 @@ from .normalize import provider_id_from_name
 logger = getLogger(__file__)
 
 
+# Map Location fields to concordance authorities
+LEGACY_CONCORDANCE_MAP = {
+    "google_places_id": "google_places",
+    "vaccinefinder_location_id": "vaccinefinder_org",
+    "vaccinespotter_location_id": "vaccinespotter_org",
+}
+
+
 def is_concordance_similar(
     source: location.NormalizedLocation,
     candidate: dict,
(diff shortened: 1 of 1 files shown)


bryanculbertson:
Not sure if I should do this or if we should clean up the data in VIAL instead.

vaccine_feed_ingest/utils/match.py
```diff
@@ -67,6 +75,17 @@ def is_concordance_similar(
 
         candidate_concordance[authority].add(value)
 
+    # Handle legacy concordance model by copying over the legacy field
+    # ids as though there was a concordance entry for them.
+    for field, authority in LEGACY_CONCORDANCE_MAP.items():
+        if not candidate_props.get(field):
+            continue
+
+        if authority in candidate_concordance:
```
eliblock, line 84: It would be interesting to check if the values matched (we could also do this as a one time query in VIAL)

eliblock (approved):
In the medium term I'd like to see us clean up these deprecated fields in VIAL (as we've already seen them cause confusion for folks working with the data downstream), but this seems like a good way to take advantage of them for matching in the meantime / until we've converted them!

vaccine_feed_ingest/utils/match.py
bryanculbertson: Think this is more important as a query in VIAL instead of buried in the matcher logs

PR #574, opened 2021-05-07
Use provider tags to exclude a match

bryanculbertson:
1)
Add the concept of provider tags to concordance e.g. `_tag_provider:rite_aid`.
These tags start with an `_` which indicate that they should be ignored for the purposed of unique matching. These tags will flow from source locations to locations when they are matched.

2) Use these tags to exclude match candidates. If the candidate has a tag for a different provider than the source then don't match it. This will allow us to be more aggressive about creating new locations that are physically close if they both have provider tags that are different.


eliblock (approved):

nschiefer (approved):

bryanculbertson:
Note that it will take two runs to see the benefit of this. One run to add the provider tags to existing locations, and then the next run can use them to create more new locations

PR #575, opened 2021-05-08
Relax phone number exclusion to only mismatch by area code

bryanculbertson:
I am trying to handle the case were these two confirmed **different** clinics have the same address, but a different phone number:
1. "The Little Clinic": https://vial-staging.calltheshots.us/admin/core/location/19179/change/
2. "Kroger": https://vial-staging.calltheshots.us/admin/core/location/18821/change/

Previously I skipped matching if phone numbers differed, but it is common for locations to have a 800 number and a location number and for different feeds to use different ones.

This relaxes the exclusion rule to only compare phone numbers with the same area code. This should hopefully still skip  merging "The Little Clinics" while still getting locations with different numbers in each feed.

### DIFF
--- a/vaccine_feed_ingest/utils/match.py
+++ b/vaccine_feed_ingest/utils/match.py
@@ -180,8 +180,8 @@ def is_phone_number_similar(
     """Compares phone numbers by area code
 
     - True if has at least one matching phone number
-    - False is mismatching phone numbers within the same area code
-    - None if no valid phone numbers to compare
+    - False if mismatching phone numbers within the same area code
+    - None if no valid phone numbers to compare, or valid phone numbers are in different area codes
     """
     candidate_props = candidate.get("properties", {})
 
--- a/vaccine_feed_ingest/utils/match.py
+++ b/vaccine_feed_ingest/utils/match.py
@@ -173,14 +173,14 @@ def is_provider_similar(
     return jellyfish.jaro_winkler(src_org, cand_org) >= threshold
 
 
-def has_matching_phone_number(
+def is_phone_number_similar(
     source: location.NormalizedLocation,
     candidate: dict,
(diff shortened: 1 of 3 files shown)


vaccine_feed_ingest/utils/match.py
```diff
@@ -220,7 +220,21 @@ def has_matching_phone_number(
         if src_phone == cand_phone:
             return True
 
-    return False
+    # Only consider phone mismatched if numbers have the same area code and do not match
+    cand_area_code = str(cand_phone.national_number)[:3]
+
+    for src_phone in src_phones:
```
eliblock, line 226: nit: should we combine this loop and the one above it?

vaccine_feed_ingest/utils/match.py
```diff
@@ -173,14 +173,14 @@ def is_provider_similar(
...
     source: location.NormalizedLocation,
     candidate: dict,
 ) -> Optional[bool]:
-    """Compares phone numbers
+    """Compares phone numbers by area code
 
     - True if has at least one matching phone number
-    - False is only mismatching phone numbers
+    - False is mismatching phone numbers within the same area code
     - None if no valid phone numbers to compare
```
eliblock, line 184: nit: 
```suggestion
    - False if mismatching phone numbers within the same area code
    - None if no valid phone numbers to compare, or valid phone numbers are in different area codes
```

eliblock (approved):
👍 this seems like a good update to the matching rules!

vaccine_feed_ingest/utils/match.py
bryanculbertson: Yeah, these could be combined. Still need to go through all of the phone numbers though to declare a mismatch. I'll do that in another PR.

PR #613, opened 2021-05-12
Common Phone Number Normalizer

cdjj:
## Notes

Adds a helper to normalize phone numbers, using the [phonenumbers](https://github.com/daviddrysdale/python-phonenumbers) library.  Prompted by @bryanculbertson 's comment in https://github.com/CAVaccineInventory/vaccine-feed-ingest/pull/593#discussion_r629751377.

I have separate changes in flight (https://github.com/cdjj/vaccine-feed-ingest/commit/39dd4cd16459cfb9e89cb3c548de84927f49e450) to convert existing normalizers to use this helper.  Data appears to be the same with both the existing normalizers, typically using ad hoc regexes, and this helper.  Key exception so far is `nc/myspot_gov`, which has data that doesn't fit the expected format, e.g. using "NNN-NNN-NNNN, option 3"; this helper would only extract the "NNN-NNN-NNNN" and ignore "option 3".  I'd welcome comments on:

1. whether it makes sense to continue that set of changes or just make this available for future normalizers; and
2. whether to skip converting `nc/myspot_gov`.

## Before Opening a PR
- [ X I ran auto-formatting: `poetry run tox -e lint-fix`


bryanculbertson:
@cdjj this is great, thank you for adding this helper.

What do you think about replacing the string "option" with "ext" before running the parser? Phonenumbers knows how to parse ext.

bryanculbertson (approved):

cdjj:
Yeah, I was thinking about that --- from a user's perspective, I doubt anyone's going to care about the distinction between an extension and an option when calling.  Working on an update and tests to do that.

cdjj:
OK, there we go.  That should pick up the bulk of what `nc/myspot_gov` matches.

bryanculbertson (approved):
This is great. Thank you for adding the options parsing.

PR #619, opened 2021-05-12
Impose a max source location record size of 15KB and fix vaccinespotter

bryanculbertson:
`vaccinespotter_org` had 65KB+ record sizes because they have every appointment time available as a huge list! This was causing import to fail because the post body was too big.

We aren't using this information anywhere so I am stripping it out of the `source.data` field in the `normalize` stage. If we want to use this data in the future we can add it back as a normalized field.

I also added the ability to exclude sites when processing so I could skip the scrappers that are broken right now.


bryanculbertson:
@eliblock I am going to merge this fix in so it goes into the next run. Feel free to post-review

PR #632, opened 2021-05-13
Phone normalization: 211, 311, additional tests

cdjj:
Handles 211 and 311, which `phonenumbers` library doesn't match.

## Before Opening a PR


vaccine_feed_ingest/utils/normalize.py
```diff
@@ -242,11 +242,20 @@ def normalize_phone(phone: Optional[str]) -> Optional[List[str]]:
...
+    phone = re.sub(r"\b211\b", "(900) 211-0000", phone)
+    phone = re.sub(r"\b311\b", "(900) 311-0000", phone)
+
     phones = []
     for match in phonenumbers.PhoneNumberMatcher(phone, "US"):
-        phones.append(
-            phonenumbers.format_number(
-                match.number, phonenumbers.PhoneNumberFormat.NATIONAL
+        if match.number.national_number == 9002110000:
+            phones.append("211")
```
eliblock, line 252: @bryanculbertson - do we expect downstream systems to be able to handle these?

eliblock (commented):
Apologies for the delay - I'm not 100% confident on how these will be handled by downstream systems (maybe they'll see the same issues that the pre-PR logic here did), so want to check with @bryanculbertson to confirm.

vaccine_feed_ingest/utils/normalize.py
bryanculbertson: @cdjj I don't think N11 codes will work with down systems. Can you add either skip these phone codes, or add these codes to the `other` field instead of to the `phone` field?

cdjj: Hmm, we'll need to change this so it returns Contact objects rather than a list of strings that then get turned into Contacts by the normalizers, but that's straightforward.

cdjj: OK, refactored; PTAL.

bryanculbertson (approved):
Thank you for figuring this out!

PR #633, opened 2021-05-13
Replace ad hoc phone normalization

cdjj:
| Key Details |
|-|
State: several |
Site: several |

## Notes
Please do **not** merge this yet.

Is there a good way to compare output before and after this patch?  Simply diff-ing the output from a normalization run with the old code and this is noisy, since most normalizers include the current timestamp in their output.  I'd really prefer to do something better than eyeballing the output, so that may mean writing some code if there isn't already some easy way to do that....

## Before Opening a PR
- [ ] I tested this using the CLI (e.g., `poetry run vaccine-feed-ingest <state>/<site>`)


bryanculbertson:
@cdjj you can try logging the previous and new phone numbers in the same log line?

Or you can do one run with the old code, and one run with the new code, and us `jq` to filter the data in `out` for to just the id and phone, and use a file diff to compare?

cdjj:
`jq` was a great suggestion, @bryanculbertson ; diff's at https://gist.github.com/cdjj/ba6513076e501520945684c7c2acd7e4.

The diffs for us/giscorps_vaccine_providers are all due to invalid phone numbers (typically typos with too many / too few characters to be valid, or invalid area codes), so the numbers wouldn't've worked in the first place; maybe something to bring up with upstream?  This also caught a neat off-by-one error (https://gist.github.com/cdjj/ba6513076e501520945684c7c2acd7e4#file-diff-ndjson-L162) in the original normalizer.

AFAICT, this is good to go.

bryanculbertson (approved):
Looks at those awesome parsed phone numbers! 🤩

Thanks for tackling this and doing the comparison. Approving and merging!

PR #635, opened 2021-05-13
Add caching framework for caching api calls during enrichment 

bryanculbertson:
Creates the concept of a cache which stores items as JSON entries in a sqllite3 DB using [diskcache](http://www.grantjenks.com/docs/diskcache).

A "api_cache" is then passed to enrichment stage and this cache can be used to cache api calls so we don't call the api everytime. Each site has its own "api_cache" so feeds can still be processed independently.

The cache is stored in GCS as a gzipped tar at each site's enrich root  (e.g. `us/vaccinespotter_org/enrich/.api_cache.tar.gz`), and is loaded at the beginning of enriching a site feed, and then stored when finished.

The usage of this cache will be defined in a future PR, but the idea is to use this like a memoize cache for api calls with a cache expire of maybe `45 +/- 15` days, so we can refresh things eventually, but not overwhelm servers

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -40,4 +40,5 @@ commands =
 deps =
   test
 commands=
+  poetry install --quiet --no-root --no-interaction
   poetry run pytest tests {posargs}
--- a/vaccine_feed_ingest/stages/ingest.py
+++ b/vaccine_feed_ingest/stages/ingest.py
@@ -12,7 +12,7 @@ from vaccine_feed_ingest_schema import location
 from vaccine_feed_ingest.utils.log import getLogger
 
 from ..utils.validation import VACCINATE_THE_STATES_BOUNDARY
-from . import enrichment, outputs, site
+from . import caching, enrichment, outputs, site
 from .common import STAGE_OUTPUT_SUFFIX, PipelineStage
 
 logger = getLogger(__file__)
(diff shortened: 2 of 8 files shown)


vaccine_feed_ingest/stages/ingest.py
```diff
@@ -340,7 +340,18 @@ def run_enrich(
             enrich_output_dir,
         )
 
-        success = enrichment.enrich_locations(enrich_input_dir, enrich_output_dir)
+        enrich_stage_dir = outputs.generate_stage_dir(
+            output_dir,
+            site_dir.parent.name,
+            site_dir.name,
+            PipelineStage.ENRICH,
+        )
```
eliblock, line 348: the relevant root directory (e.g., `out/al/arcgis/enriched`) will not exist the first time this is run, but we assume it does exist within `caching` (we try to open write to `out/al/arcgis/enriched/.api_cache.tar.gz`).

Should we create the relevant directory first here?

tox.ini
```diff
@@ -40,4 +40,5 @@ commands =
 deps =
   test
 commands=
+  poetry install --quiet --no-root --no-interaction
```
eliblock, line 43: Why is this line newly needed? Shouldn't `tox` handle installing the dependencies out of the box?

bryanculbertson: This is to handle the case where people add new requirements and other devs don't recreate their tox envs.

This is a no-op if the requirements are up-to-date

vaccine_feed_ingest/stages/ingest.py
bryanculbertson: Good point, I'll fix this

bryanculbertson: Fixed

PR #638, opened 2021-05-14
fix: stable ids for il/sf_sites

eliblock:
Presently the ids for `il/sf_sites` change on each run of the pipeline, so we disabled the site. Manually create a stable id instead.


bryanculbertson (approved):
Note this will break if a site does not have a name or a street address

eliblock:
Good callout! This data thusfar seems to regularly include those fields, but I bet we'll be looking at this again in the future 🙃.

PR #644, opened 2021-05-15
Support for adding placekey concordance links from placekey.io during enrichment

bryanculbertson:
Support for adding data from external APIs during enrich stage. External API calls are cached for 35-50 days so we don't overwhelm the external service.

Placekey is free, but slow, so we may need to wait to enable this until we optimize out load-to-vial runtime

You need to get an api key from https://dev.placekey.io/ in order for this to work

e.g.
```
vaccine-feed-ingest enrich ca/sf_gov --enrich-apis=placekey --placekey-apy-key=***
```

### DIFF
--- a/vaccine_feed_ingest/apis/common.py
+++ b/vaccine_feed_ingest/apis/common.py
@@ -1,0 +1,52 @@
+"""Common API tooling"""
+import hashlib
+import random
+from typing import Any, Optional, Sequence
+
+import diskcache
+
+DEFAULT_EXPIRE_SECONDS = 45 * 24 * 60 * 60  # 40 days
+DEFAULT_EXPIRE_JIGGLE_PERCENT = 0.10  # +/- 4 days
+
+
+def calculate_cache_key(name: str, args: Sequence[str]) -> str:
+    """Convert a cache key from a sequence of str values"""
+    hasher = hashlib.md5()
+
+    for value in args:
+        hasher.update(value.encode())
+
+    args_hash = hasher.hexdigest()
+
+    return f"{name}:{args_hash}"
+
+
+class CachedAPI:
+    """API for calling placekey that checks the cache first"""
+
+    def __init__(
+        self,
+        api_cache: diskcache.Cache,
+        expire_secs: Optional[float] = None,
+        expire_jiggle_percent: Optional[float] = None,
+    ):
+        self.api_cache = api_cache
+
+        if expire_secs is None:
+            expire_secs = DEFAULT_EXPIRE_SECONDS
+
+        if expire_jiggle_percent is None:
+            expire_jiggle_percent = DEFAULT_EXPIRE_JIGGLE_PERCENT
+
+        self._expire_min_secs = expire_secs - expire_secs * expire_jiggle_percent
+        self._expire_max_secs = expire_secs + expire_secs * expire_jiggle_percent
+
+    def _calculate_expire(self) -> float:
+        """Calculate a random expire number between the range.
+
+        This is so we don't expire all of the locations at the same time.
+        """
+        return random.uniform(self._expire_min_secs, self._expire_max_secs)
+
+    def set_with_expire(self, key: str, value: Any) -> bool:
+        return self.api_cache.set(key, value, expire=self._calculate_expire())
--- a/vaccine_feed_ingest/apis/placekey.py
+++ b/vaccine_feed_ingest/apis/placekey.py
@@ -1,0 +1,77 @@
+from typing import Optional
+
+import diskcache
+import placekey.api
+
+from ..utils.log import getLogger
+from .common import CachedAPI, calculate_cache_key
+
+logger = getLogger(__file__)
+
+
+class PlacekeyAPI(CachedAPI):
+    """API for calling placekey that checks the cache first"""
+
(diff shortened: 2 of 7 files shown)


vaccine_feed_ingest/apis/common.py
```diff
@@ -0,0 +1,52 @@
...
+        if expire_jiggle_percent is None:
+            expire_jiggle_percent = DEFAULT_EXPIRE_JIGGLE_PERCENT
+
+        self._expire_min_secs = expire_secs - expire_secs * expire_jiggle_percent
+        self._expire_max_secs = expire_secs + expire_secs * expire_jiggle_percent
+
+    def _calculate_expire(self) -> float:
+        """Calculate a random expire number between the range.
+
+        This is so we don't expire all of the locations at the same time.
```
eliblock, line 47: 👍 good planning!

vaccine_feed_ingest/apis/placekey.py
```diff
@@ -0,0 +1,77 @@
...
+            strict_address_match=strict_address_match,
+            strict_name_match=strict_name_match,
+        )
+
+        if not response:
+            return None
+
+        if "error" in response:
+            logger.info("Failed to add placekey because: %s", response["error"])
+            self.set_with_expire(cache_key, {"error": response["error"]})
```
eliblock, line 70: Do we want to cache errors? How can we separate intermittent failures from persistent ones?

vaccine_feed_ingest/apis/placekey.py
```diff
@@ -0,0 +1,77 @@
...
+                postal_code,
+                iso_country_code,
+                str(strict_address_match),
+                str(strict_name_match),
+            ],
+        )
+
+        cache_response = self.api_cache.get(cache_key)
+        if cache_response:
+            return cache_response.get("placekey")
```
eliblock, line 50: Should we check if the cached value was an error (and return `None` if so)?

vaccine_feed_ingest/cli.py
```diff
@@ -89,6 +89,27 @@ def _stages_option() -> Callable:
...
+        ),
+    )
+
+
+def _placekey_apikey_option() -> Callable:
+    return click.option(
+        "--placekey-apikey",
+        "placekey_apikey",
+        type=str,
+        default=lambda: os.environ.get("PLACEKEY_APIKEY", ""),
```
eliblock, line 109: 👍 I can add this to the worker environment

vaccine_feed_ingest/stages/enrichment.py
```diff
@@ -23,12 +24,25 @@ def enrich_locations(
...
 
+    if enrich_apis is None:
+        enrich_apis = set()
+
+    placekey_api = None
+    if "placekey" in enrich_apis:
+        if api_cache is not None and placekey_apikey:
+            placekey_api = PlacekeyAPI(api_cache, placekey_apikey)
+        else:
+            logger.info("Skipping placekey because placekey api is not configured")
```
eliblock, line 41: nit: should this be `logger.warning` (or higher?)

vaccine_feed_ingest/stages/enrichment.py
```diff
@@ -201,6 +221,47 @@ def _add_provider_tag(loc: location.NormalizedLocation) -> None:
...
+) -> None:
+    """Add placekey concordance to location if we can"""
+    if placekey_api is None:
+        return
+
+    if not loc.location:
+        return
+
+    if not _valid_address(loc):
+        return
```
eliblock, line 236: We call this here and in the function which calls this function. Can/should we optimize?

vaccine_feed_ingest/apis/placekey.py
bryanculbertson: The `.get` will return `None` if there was a cached `error`

vaccine_feed_ingest/apis/placekey.py
bryanculbertson: I couldn't figure out how to distinguish  intermittent failures from persistent ones :-/

All you get back is a dict with an "error" field with human readable text describing the error, there isn't an error code. I could try to parse the error text, but that seems brittle

bryanculbertson: Examples of errors are, paraphrased:
- "error": "missing required field city"
- "error": "could not find place for specified address"

vaccine_feed_ingest/stages/enrichment.py
bryanculbertson: Yep, I upgraded to error because this means you requested that the placekey api be used, but we couldn't call it.

vaccine_feed_ingest/apis/placekey.py
bryanculbertson: I will explicitly return `None` with there is an error though

vaccine_feed_ingest/stages/enrichment.py
bryanculbertson: I decided that answer is not to optimize because:
1. The check here is to make sure we can all placekey
2. The check later is to avoid importing locations that VIAL cannot handle.

In the future we want to remove check 2 to load locations with invalid addresses to VIAL, but we can't remove check 1.

PR #660, opened 2021-05-17
fix: vial `id` at root, not in properties

eliblock:
On 2021-05-14, `VIAL` updated its `nlgeojson` response format per feedback from a partner.

Among other updates, the `id` field which was previously available both at `.id` and `.properties.id` was consolidated to `.id` only.

Update matching logic to use the consolidated field.

NOTE: it is possible that other schema changes may be breaking to some logic herein. I did not attempt to re-validate existing matching rules against the schema changes, only to fix the issue blocking streaming locations from VIAL.

### DIFF
--- a/vaccine_feed_ingest/vial.py
+++ b/vaccine_feed_ingest/vial.py
@@ -73,6 +73,10 @@ def import_source_locations(
     import_batch_size: int = 500,
 ) -> None:
     """Import source locations"""
+    path_and_query = f"/api/importSourceLocations?import_run_id={import_run_id}"
+    logger.info("Contacting VIAL: POST %s", path_and_query)
+
+    batches = 0
     for import_locations_batch in misc.batch(import_locations, import_batch_size):
         encoded_ndjson = "\n".join(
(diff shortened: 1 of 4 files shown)


vaccine_feed_ingest/vial.py
```diff
@@ -169,16 +187,23 @@ def search_source_locations(
 
     query = urllib.parse.urlencode(params)
 
-    resp = vial_http.request(
-        "GET", f"/api/searchSourceLocations?{query}", preload_content=False
-    )
+    path_and_query = f"/api/searchLocations?{query}"
```
bryanculbertson, line 190: This is wrong! We need to grab existing matches from `source_locations` table, not existing locations

bryanculbertson (changes requested):
it all looks good, but there is a bad copy and paste that breaks matching

vaccine_feed_ingest/vial.py
eliblock: *great* catch, thank you.

bryanculbertson (approved):

PR #668, opened 2021-05-20
Use orjson everywhere for performance reasons

bryanculbertson:


bryanculbertson:
@eliblock I am going to merge this without review, as well as the following PR with `content_hash`. Please review in post.

PR #674, opened 2021-05-21
Add placekeys with bulk API which is way faster

bryanculbertson:
Collect all of the locations and pas them all at once to bulk api. Their library will batch into groups of 100, so I don't need to.

I think we can enable place keys after this lands!

### DIFF
--- a/vaccine_feed_ingest/apis/placekey.py
+++ b/vaccine_feed_ingest/apis/placekey.py
@@ -57,6 +57,10 @@ class PlacekeyAPI(CachedAPI):
     ) -> Dict[str, Optional[str]]:
         result: Dict[str, Optional[str]] = {}
 
+        places = []
+        cache_keys = {}
+
+        # Load valid records from cache
         for record_id, record in records.items():
             if not (latitude := record.get("latitude")):
                 continue
@@ -75,7 +79,7 @@ class PlacekeyAPI(CachedAPI):
             if not (iso_country_code := record.get("iso_country_code", "US")):
                 continue
 
-            cache_key = calculate_cache_key(
+            cache_keys[record_id] = calculate_cache_key(
                 "placekey",
                 [
                     f"{latitude:.5f}",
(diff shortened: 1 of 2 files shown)


vaccine_feed_ingest/apis/placekey.py
```diff
@@ -29,53 +29,105 @@ def lookup_placekey(
...
+    def lookup_placekeys(
+        self,
+        records: Dict[str, dict],
+        strict_address_match: bool = False,
+        strict_name_match: bool = False,
+    ) -> Dict[str, Optional[str]]:
+        result: Dict[str, Optional[str]] = {}
+
+        for record_id, record in records.items():
+            if not (latitude := record.get("latitude")):
```
eliblock, line 65: Should we log in the event that any of these fail?

vaccine_feed_ingest/apis/placekey.py
```diff
@@ -29,53 +29,105 @@ def lookup_placekey(
...
+                strict_address_match=strict_address_match,
+                strict_name_match=strict_name_match,
+            )
+
+            if not response:
+                result[record_id] = None
+                continue
+
+            if "error" in response:
+                logger.info("Failed to add placekey because: %s", response["error"])
```
eliblock, line 139: There are several other failure cases. Should we log before `continue`-ing (or `return`-ing) in them?

vaccine_feed_ingest/apis/placekey.py
```diff
@@ -91,43 +95,58 @@ def lookup_placekeys(
...
             if not response:
-                result[record_id] = None
                 continue
 
             if "error" in response:
                 logger.info("Failed to add placekey because: %s", response["error"])
-                self.set_with_expire(cache_key, {"error": response["error"]})
-                result[record_id] = None
+                self.set_with_expire(
+                    cache_keys[record_id], {"error": response["error"]}
```
eliblock, line 141: Should we set a _shorter_ cache time for errors, or do we expect them to be persistent?

From a local run it looks like the errors I see are all related to the source data (e.g., bad address, which will be handled by the cache key).

eliblock (approved):

vaccine_feed_ingest/apis/placekey.py
bryanculbertson: I don't want a lower cache time unless it is temporary error on the server side, which I don't really have a hook to know when there is a temporary error.

I definetly want to cache when the address is bad so we don't continue to call placekey for bad addresses.

vaccine_feed_ingest/apis/placekey.py
bryanculbertson: Added

vaccine_feed_ingest/apis/placekey.py
bryanculbertson: added

PR #600, opened 2021-05-11
Normalize Physicians' Immediate Care for US

rjackson90:
# 

| Key Details |
|-|
| Resolves #433  |
State: us |
Site: physicians_immediate |

## Notes
- This isn't _quite_ complete. I haven't yet normalized store hours for the handful of sites that have them. It may be a few days before I get more time to work on this, so I wanted to get a working PR opened up rather than wait for it to be perfect. 
- I had to alter the parser to add some row ids which existed in the source HTML. I didn't realize those would be helpful when I wrote the parser. 
- ~~I'm not sure if the way I handled addresses is correct. I'm nervous that leaving the full address (including state and ZIP where available) in the `street1` field might break something downstream.~~ **FIXED (2021-05-16)**

## Data sample (Original)

```json
{"id": "us_physicians_immediate:31509_144", "name": "Niles", "address": {"street1": "8630 W. Golf Road, Niles, IL 60714", "street2": null, "city": null, "state": "IL", "zip": "60714"}, "location": null, "contact": null, "languages": null, "opening_dates": null, "opening_hours": null, "availability": null, "inventory": [{"vaccine": "moderna", "supply_level": "in_stock"}], "access": null, "parent_organization": null, "links": null, "notes": null, "active": null, "source": {"source": "us_physicians_immediate", "id": "144", "fetched_from_uri": "https://physiciansimmediatecare.com/covid-19-vaccination-locations/", "fetched_at": "2021-05-11T06:59:54.007079", "published_at": null, "data": {"clinic": "Niles", "address": "8630 W. Golf Road, Niles, IL 60714", "slots": "10", "type": "Moderna", "row_id": "144", "table_id": "31509"}}}
[...]

PR #677, opened 2021-05-21
parse street addres from GISCorps dataset using usaddress library

MoralCode:
| Key Details |
|-|
State: us |
Site: giscorps_vaccine_providers |

## Notes
update parsing to attempt to parse addresses.

There seem to still be many addresses that are ambiguous, but this gets the majority that are fully-qualified

## Before Opening a PR
- [ X] I tested this using the CLI (e.g., `poetry run vaccine-feed-ingest <state>/<site>`)
- [ X] I ran auto-formatting: `poetry run tox -e lint-fix`

### DIFF
--- a/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
+++ b/vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
@@ -261,10 +261,46 @@ def normalize_state_name(name: str) -> str:
 
 
 def apply_address_fixups(address: OrderedDict[str, str]) -> OrderedDict[str, str]:
+
+    if "PlaceName" in address and "StateName" in address:
+        problem_dakotas = ["Valley City, North", "Williston North", "Belle Fourche, South"]
+        if address["PlaceName"] in problem_dakotas and address["StateName"] == "Dakota":
+            pl_old = address["PlaceName"]
+            address["PlaceName"] = pl_old[:-5].strip()
+            address["StateName"] = pl_old[-5:] + " Dakota"
+
     if "StateName" in address:
         state = address["StateName"]
+
+        if state == "ND North Dakota":
+            state = "North Dakota"
+        elif state == "Mich.":
+            state = "Michigan"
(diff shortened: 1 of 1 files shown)


eliblock (changes requested):
@MoralCode - as discussed offline, it looks like this update currently prevents the normalizer from running. Could you take another look?

Be sure to run `poetry run vaccine-feed-ingest all-stages us/giscorps_vaccine_providers` and check the output!

Additionally, you may want to rebase on top of `main`. In #600 we added another user of `usaddress`, and they handled dependency import! (The code from that PR might also help debug this normalizer, though I haven't looked closely).

MoralCode:
@eliblock seems to run fine for me. what problem were you seeing where it was stopping the normalizer from running?

also what do you mean "they handled dependency import"? `usaddress` is imported the same in both PR's

MoralCode:
also, would it help to hold off and integrate this with #691?

eliblock:
@MoralCode great suggestion - let's build this on top of #691!

MoralCode:
@eliblock rebased on top of the latest main branch and updated to utilize the utility parser from the mentioned PR. Should now be good to go

vaccine_feed_ingest/runners/us/giscorps_vaccine_providers/normalize.py
```diff
@@ -175,29 +185,188 @@ def try_get_lat_long(site):
     return location
 
 
-def _get_normalized_location(site: dict, timestamp: str) -> schema.NormalizedLocation:
+def normalize_state_name(name: str) -> str:
+
+    if name is None:
+        return name
+
+    us_state_abbrev = {
```
bryanculbertson, line 193: For a followup PR, you can convert this to use the `us` library.

e.g.
```
import us
us.states.lookup('maryland')
```

bryanculbertson (approved):

bryanculbertson:
Approving and merging. There is some followup work to use `us` library if you want to simplify your code somewhat

PR #692, opened 2021-05-27
Fixed installation script for mac.

MaxBrady:
* `set -e` was causing maybe_install to fail when a package was not installed. `which $1` was returning an error when the package was not installed.
* Did not make any changes to Linux install function.
* Getting an error installing poppler:
```
==> Installing gdal dependency: poppler
Error: Cannot install poppler because conflicting formulae are installed.
pdftohtml: because poppler, pdftohtml, pdf2image, and xpdf install conflicting executables
```
This error is unrelated to the script as it is reproducible using manual install steps. I have not been able to resolve this error yet.

* Unable to run lint or test because installation is broken.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,9 +1,5 @@
 #!/usr/bin/env bash
 
-# exit when a command fails instead of blindly blundering forward
-set -e
-# treat unset variables as an error and exit immediately
-set -u
 # don't hide exit codes when pipeline output to another command
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


obra (approved):
Looks fine to me. I'd *rather* we fixed maybe_install but don't have the cycles to do that myself, so +1 to merge.

setup.sh
```diff
@@ -1,9 +1,5 @@
 #!/usr/bin/env bash
 
-# exit when a command fails instead of blindly blundering forward
-set -e
-# treat unset variables as an error and exit immediately
-set -u
```
bryanculbertson, line 6: I think we should keep `set -u` and `set -e` at the top, and then disable them just for the part that we handle errors

e.g. in maybe_install
```
set +e
$3
set -e
```

bryanculbertson (changes requested):

eliblock:
Closing all prs: https://blog.vaccinateca.com/vaccinate-the-states-is-winding-down

PR #705, opened 2021-06-02
chore: add Sentry

Vaccinebot:
Configure `sentry-sdk` when the `SENTRY_DSN` environment variable is set. This variable should be set to the proper endpoint for our Sentry project (e.g., `https://01231:a1232@b1233.ingest.sentry.io/1234`). This value can be pulled from [the sentry docs](https://docs.sentry.io/clients/python/) if you are logged in - be sure to click the green downward arrow in one of the examples to choose the proper project from the dropdown.

Sentry will automatically pick up on all `logger.error` messages (or actual errors) when it is enabled. Additional configuration (or sending some things to Sentry directly) is possible if desired.

### DIFF
--- a/vaccine_feed_ingest/cli.py
+++ b/vaccine_feed_ingest/cli.py
@@ -11,6 +11,7 @@ from typing import Callable, Collection, Dict, Optional, Sequence
 import click
 import dotenv
 import pathy
+import sentry_sdk
 
 from .stages import caching, common, ingest, load, site
 
@@ -246,6 +247,16 @@ def _import_batch_size_option() -> Callable:
 def cli():
     """Run vaccine-feed-ingest commands"""
     dotenv.load_dotenv()
+
+    sentry_path = os.environ.get("SENTRY_DSN")
+    if sentry_path:
+        sentry_sdk.init(
+            dsn=sentry_path,
+            # 1.0 would capture 100% of transactions for performance monitoring.
+            traces_sample_rate=0.0,
+        )
+    else:
+        print("Sentry disabled (no config provided).")
 
 
 @cli.command()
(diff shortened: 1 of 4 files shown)


vaccine_feed_ingest/stages/ingest.py
```diff
@@ -152,7 +157,12 @@ def run_parse(
         except CalledProcessError as e:
             if fail_on_runner_error:
                 raise e
-            logger.error("Subprocess errored, stage will be skipped: %s", e)
+            logger.error(
+                "Subprocess %s/%s errored, stage will be skipped: %s",
+                site_dir.parent.name,
+                site_dir.name,
+                e,
```
bryanculbertson, line 164: Not quite sure how sentry works, but the python standard is to add `exc_info=True` if you want the exception in the log statement

vaccine_feed_ingest/cli.py
```diff
@@ -247,6 +248,16 @@ def cli():
     """Run vaccine-feed-ingest commands"""
     dotenv.load_dotenv()
 
+    sentry_path = os.environ.get("SENTRY_DSN")
+    if sentry_path:
```
bryanculbertson, line 252: Can you add a ENABLE_SENTRY env bar too?

I like the ability to disable a service if needed without needing to remove the configuration for it

vaccine_feed_ingest/cli.py
```diff
@@ -247,6 +248,16 @@ def cli():
...
+        sentry_sdk.init(
+            dsn=sentry_path,
+            # 1.0 would capture 100% of transactions for performance monitoring.
+            traces_sample_rate=0.0,
+        )
+    else:
+        print("Sentry disabled (no config provided).")
```
bryanculbertson, line 259: This should be a log statement not a print.

vaccine_feed_ingest/cli.py
bryanculbertson: It also means you can log higher if enable sentry is true, but the config is missing

PR #765, opened 2021-06-07
FYI Skip VTRCKS for matching purposes

bryanculbertson:
VTRCKS will assign the same id to different locations, so do not trust it for matching


bryanculbertson:
Merging without review because I don't want the bad matches to go in

PR #774, opened 2021-06-09
Option to limit number of records to import for a site

bryanculbertson:
Add an option to limit the number of records to import with `--import-limit`

This is useful for testing import without needing to match all records first.

### DIFF
--- a/vaccine_feed_ingest/vial.py
+++ b/vaccine_feed_ingest/vial.py
@@ -89,12 +89,16 @@ def import_source_locations(
             ]
         )
 
-        rsp = vial_http.request(
-            "POST",
-            path_and_query,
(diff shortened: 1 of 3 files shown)


vaccine_feed_ingest/vial.py
```diff
@@ -89,12 +89,16 @@ def import_source_locations(
...
-        )
+        try:
+            rsp = vial_http.request(
+                "POST",
+                path_and_query,
+                headers={**vial_http.headers, "Content-Type": "application/x-ndjson"},
+                body=encoded_ndjson,
+            )
+        except urllib3.exceptions.ReadTimeoutError:
+            logger.error("Timeout while importing: %s", encoded_ndjson[:100])
```
eliblock, line 100: Thank you!

eliblock (approved):

PR #779, opened 2021-06-09
disable az/pinal_ph_vaccinelocations_gov 

MoralCode:
Data seems to have disappeared as the [original page](https://www.pinalcountyaz.gov/publichealth/CoronaVirus/Pages/vaccinelocations.aspx) no longer has a table and simply links to the state DOH list


MoralCode:
Funnily enough i did this because i saw another scraper had been disabled this way when i was looking at something else the other day. seems like a neat little hack

PR #811, opened 2021-06-30
rename ri prepmod so its clear that it comes from RI

MoralCode:
this is a possibly disruptive change to hopefully make errors involving this scraper easier to track down

if no other scrapers are named `prepmod` then this is probably not worth the disruption to fix

### DIFF
--- a/vaccine_feed_ingest/runners/ri/prepmod/fetch.yml
+++ b/vaccine_feed_ingest/runners/ri/prepmod/fetch.yml
@@ -1,5 +1,5 @@
 ---
 state: ri
-site: prepmod
+site: ri_prepmod
 parser: prepmod
 url: https://www.vaccinateri.org/


MoralCode:
normalize and fetch yml files still need changing

vaccine_feed_ingest/runners/ri/prepmod/fetch.yml
```diff
@@ -1,5 +1,5 @@
 ---
 state: ri
-site: prepmod
+site: ri_prepmod
```
bryanculbertson, line 3: This will change the ids for all of these locations, right? If so then it is too late to make this change.

MoralCode: yeah :/ thats why i was concerned about how breaking this change was.

on one hand if im combining the duplicate prepmod locations as part of #814 the ID's will change anyway, on the other hand, if this is the only scraper with the generic `prepmod` id, then theres no confusion and this issue is more pedantic than an actual problem

eliblock:
Closing all prs: https://blog.vaccinateca.com/vaccinate-the-states-is-winding-down

PR #817, opened 2021-07-01
Mn gov url fixups

MoralCode:
## Notes
fix mn_gov url parsing to handle some additional typos (primarily a url with only one forward slash)

## Before Opening a PR

### DIFF
--- a/vaccine_feed_ingest/runners/mn/gov/normalize.py
+++ b/vaccine_feed_ingest/runners/mn/gov/normalize.py
@@ -54,7 +54,7 @@ def _cleanup_url(url):
         return None
 
     url = re.sub(r"^(https?)\/:([^\/:].*)", r"\g<1>://\g<2>", url)
-    url = re.sub(r"^https:/t", "https://t", url)  # fix typos
+    url = re.sub(r"^(https?:\/)([^\/].*)", r"\g<1>/\g<2>", url)
 
     return url
 
--- a/vaccine_feed_ingest/runners/mn/gov/normalize.py
+++ b/vaccine_feed_ingest/runners/mn/gov/normalize.py
@@ -48,10 +48,6 @@ def _get_id(site: dict) -> str:
 def _cleanup_url(url):
     if not url:
         return None
-    # if "@" in url:
-    #     # Some of these are email addresses.
-    #     # Skipping those for now.
-    #     return None
     if " " in url.strip():
         # url contains spaces after stripping the leading and trailing ones
         # it must be an invalid URL
@@ -59,9 +55,6 @@ def _cleanup_url(url):
 
     url = re.sub(r"^https/:", "https://", url)  # fix typos
     url = re.sub(r"^https:/t", "https://t", url)  # fix typos
-
-    # if not url.startswith("http"):
-    #     url = "http://" + url
 
     return url
 
--- a/vaccine_feed_ingest/runners/mn/gov/normalize.py
+++ b/vaccine_feed_ingest/runners/mn/gov/normalize.py
@@ -12,6 +12,7 @@ from pydantic import ValidationError
 from vaccine_feed_ingest_schema import location as schema
 
 from vaccine_feed_ingest.utils.log import getLogger
+from vaccine_feed_ingest.utils.normalize import normalize_url
 
 logger = getLogger(__file__)
 
@@ -43,6 +44,25 @@ def _get_id(site: dict) -> str:
 
     return f"{runner}_{site_name}:{arcgis}_{layer}_{data_id}"
 
+def _cleanup_url(url):
+    if not url:
+        return None
+    # if "@" in url:
+    #     # Some of these are email addresses.
+    #     # Skipping those for now.
+    #     return None
+    if url == "COVID-19 Vaccine Information - Hennepin Healthcare":
+        return None
+    print(url)
+    
+    url = re.sub(r"^https/:", "https://", url)  # fix typos
+    url = re.sub(r"^https:/t", "https://t", url)  # fix typos
+
+    # if not url.startswith("http"):
+    #     url = "http://" + url
+
+    return url
+
 
 def _get_contacts(site: dict) -> Optional[List[schema.Contact]]:
     contacts = []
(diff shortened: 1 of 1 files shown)


vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
     return f"{runner}_{site_name}:{arcgis}_{layer}_{data_id}"
 
 
+def _cleanup_url(url):
+    if not url:
+        return None
+    # if "@" in url:
```
bryanculbertson, line 51: Why did you comment out the email check?

vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
...
+    #     # Some of these are email addresses.
+    #     # Skipping those for now.
+    #     return None
+    if url == "COVID-19 Vaccine Information - Hennepin Healthcare":
+        return None
+    print(url)
```
bryanculbertson, line 57: Remove the debugging print statement

vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
...
 
 
+def _cleanup_url(url):
+    if not url:
+        return None
+    # if "@" in url:
+    #     # Some of these are email addresses.
+    #     # Skipping those for now.
+    #     return None
+    if url == "COVID-19 Vaccine Information - Hennepin Healthcare":
```
bryanculbertson, line 55: Perhaps a more generic rule would be to strip leading and trailing spaces, and then skip entries that still have spaces (since the can't be a url)

vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
...
+        return None
+    print(url)
+
+    url = re.sub(r"^https/:", "https://", url)  # fix typos
+    url = re.sub(r"^https:/t", "https://t", url)  # fix typos
+
+    # if not url.startswith("http"):
```
bryanculbertson, line 62: This was probably fixing urls like` www.vaccine.org` so we should keep that?

vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
...
+    # if "@" in url:
+    #     # Some of these are email addresses.
+    #     # Skipping those for now.
+    #     return None
+    if url == "COVID-19 Vaccine Information - Hennepin Healthcare":
+        return None
+    print(url)
+
+    url = re.sub(r"^https/:", "https://", url)  # fix typos
+    url = re.sub(r"^https:/t", "https://t", url)  # fix typos
```
bryanculbertson, line 60: Why is the `t` in regex, that seems very specific. Perhaps changes this to be any non-slash character?

vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
...
-        url = re.sub(r"^https/:", "https://", url)  # fix typos
-
-        if not url.startswith("http"):
-            url = "http://" + url
-
-        return url
-
-    if url := cleanup_url(site["attributes"]["URL"]):
+    if url := normalize_url(_cleanup_url(site["attributes"]["URL"])):
+        print(url)
```
bryanculbertson, line 75: Remove debugging print

bryanculbertson (changes requested):

vaccine_feed_ingest/runners/mn/gov/normalize.py
MoralCode: ah, thanks for catching that

vaccine_feed_ingest/runners/mn/gov/normalize.py
MoralCode: i essentially removed this whole fixup function by comments in an attempt to convert it to the utility URL normalizer, and then i selectively re-added the checks (and added more checks) to fix the issues that the util function doesnt handle.

in effect this is also partly a migration to using the common url normalizer

vaccine_feed_ingest/runners/mn/gov/normalize.py
MoralCode: i suspect that the common utility normalizer handles these cases already

vaccine_feed_ingest/runners/mn/gov/normalize.py
MoralCode: the t is included because otherwise it was otherwise mangling legitimate urls. i.e. it would take a valid `https://` and make it `https:///`

vaccine_feed_ingest/runners/mn/gov/normalize.py
MoralCode: good catch! Do you think this would be useful to make into a shared utility function for cleaning up urls?

vaccine_feed_ingest/runners/mn/gov/normalize.py
bryanculbertson: Ah, I see, then just remove this commented out code because it isn't relevant here anymore.

vaccine_feed_ingest/runners/mn/gov/normalize.py
bryanculbertson: Then just remove this commented out code

vaccine_feed_ingest/runners/mn/gov/normalize.py
bryanculbertson:
```suggestion
    url = re.sub(r"^(https?:\/)([^\/].*)", r"\g<1>/\g<2>", url)
```

I see, this regex is more generic and handles he cases:
1. Handle http or https
2. Escapes `/` as is the std
3. Matches any url missing `//` not just one that starts with `t`

vaccine_feed_ingest/runners/mn/gov/normalize.py
```diff
@@ -44,29 +45,34 @@ def _get_id(site: dict) -> str:
...
+    # if "@" in url:
+    #     # Some of these are email addresses.
+    #     # Skipping those for now.
+    #     return None
+    if " " in url.strip():
+        # url contains spaces after stripping the leading and trailing ones
+        # it must be an invalid URL
+        return None
+
+    url = re.sub(r"^https/:", "https://", url)  # fix typos
```
bryanculbertson, line 60:
```suggestion
    url = re.sub(r"^(https?)\/:([^\/:].*)", r"\g<1>://\g<2>", url)
```

vaccine_feed_ingest/runners/mn/gov/normalize.py
bryanculbertson: Yes

MoralCode: aight will do

bryanculbertson (approved):

